Issue #1, opened 2019-01-31
Kotlin extension runtime and annotation processor for extension specialization


JakeWharton:
Like demoed in the talk

ok3141:
Maybe it's not a right place for questions, but can you explain why do we need Dagger with reflection again?

JakeWharton:
@ok3141 See https://jakewharton.com/helping-dagger-help-you/

PR #2, opened 2019-01-31
Start Kotlin runtime

JakeWharton:
Refs #1


reflect-kotlin/src/main/java/dagger/dagger.kt
```diff
@@ -0,0 +1,8 @@
+@file:Suppress("NOTHING_TO_INLINE")
+
+package dagger
+
+import kotlin.reflect.KClass
+
+inline fun <C : Any> KClass<C>.create(): C = Dagger.create(java)
+inline fun <B : Any> KClass<B>.builder(): B = Dagger.builder(java)
```
TWiStErRob: this "pollutes" the global namespace, would there be a way to do `inline fun <@Component.Builder B : Any> KClass<B>`? (even if that needs dagger contributions)

reflect-kotlin/src/test/java/dagger/DaggerTest.kt
```diff
@@ -0,0 +1,14 @@
...
+import org.junit.Assert.assertNotNull
+import org.junit.Test
+
+class DaggerTest {
+  @Test fun create() {
+    assertNotNull(NoBuilderComponent::class.create())
+  }
+
+  @Test fun builder() {
+    assertNotNull(BuilderComponent.Builder::class.builder())
```
TWiStErRob: since there are no dependencies `BuilderComponent::class.create()` should also work?

PR #8, opened 2019-02-03
Add support for ordered @Inject in class hierarchies

TWiStErRob:
Re-work of https://github.com/JakeWharton/SdkSearch/pull/144 on top of current master addressing code review comments as well, and moving test as an integration test.


integration-tests/src/main/java/com/example/MemberInjectionOrder.java
```diff
@@ -0,0 +1,54 @@
...
+import java.util.LinkedList;
+import java.util.List;
+import javax.inject.Inject;
+
+@Component(modules = MemberInjectionOrder.Module1.class)
+interface MemberInjectionOrder {
+  void inject(SubType target);
+
+  abstract class Base {
+    List<String> calls = new LinkedList<>();
```
JakeWharton:
```suggestion
    List<String> calls = new ArrayList<>();
```

integration-tests/src/test/java/com/example/IntegrationTest.java
```diff
@@ -142,6 +142,27 @@
...
+   * [@Inject] Constructors are injected first, followed by fields, and then methods. Fields and methods in
+   * superclasses are injected before those in subclasses.
+   *
+   * @see javax.inject.Inject
+   */
+  @Test public void memberInjectionOrder() {
+    MemberInjectionOrder component = backend.create(MemberInjectionOrder.class);
+    MemberInjectionOrder.SubType target = new MemberInjectionOrder.SubType();
+    component.inject(target);
+    target.calls.add("injected: " + target);
```
JakeWharton: I don't think you need this or the booleans to track method calls. The logs already contain that information. All we need to see is constructor, base method, and sub method which will inform the field injection order based on the output.

integration-tests/src/test/java/com/example/IntegrationTest.java
```diff
@@ -142,6 +142,27 @@
     assertThat(target.subtypeCalled).isTrue();
   }
 
+  /**
+   * [@Inject] Constructors are injected first, followed by fields, and then methods. Fields and methods in
```
JakeWharton: `{@code Inject @Inject}`, although I don't think Javadoc is useful on tests. I'd rather see this as a line comment inside near the assertion.

integration-tests/src/test/java/com/example/IntegrationTest.java
TWiStErRob: good point about booleans, came over with migration; re "injected", wanted to see `subCalled` being true, but that's just testing the fixture now that I look at it fresh

TWiStErRob:
Rebased and fixed comments, see [force-push](https://github.com/JakeWharton/dagger-reflect/compare/83e8cfece985376ac712a391198c13423e44d623..054d2091f1da69de5e5b1c78dcff411476a43fda) for details, again `ArrayDeque` because `addFirst` and errorprone.

PR #13, opened 2019-02-04
Allow @Modules to be interfaces

TWiStErRob:
> getSuperclass(): If this Class represents either the Object class, an interface, a primitive type, or void, then null is returned.


TWiStErRob:
Superseded by #15

PR #15, opened 2019-02-04
Allow @Modules to be interfaces

JakeWharton:
Closes #13

### DIFF
--- a/integration-tests/src/main/java/com/example/ModuleClassHierarchy.java
+++ b/integration-tests/src/main/java/com/example/ModuleClassHierarchy.java
@@ -1,0 +1,23 @@
+package com.example;
+
+import dagger.Binds;
+import dagger.Component;
+import dagger.Module;
+import dagger.Provides;
+
+@Component(modules = ModuleClassHierarchy.Module1.class)
+interface ModuleClassHierarchy {
+  CharSequence string();
+
+  @Module
+  abstract class Module1 extends BaseModule {
+    @Provides static String string() {
+      return "foo";
+    }
+  }
+
+  @Module
+  abstract class BaseModule {
+    @Binds abstract CharSequence charSequence(String foo);
+  }
+}
--- a/integration-tests/src/test/java/com/example/IntegrationTest.java
+++ b/integration-tests/src/test/java/com/example/IntegrationTest.java
@@ -256,6 +256,43 @@ public final class IntegrationTest {
     assertThat(values.get("1").get()).isEqualTo("one");
   }
 
+  @Test public void moduleClass() {
+    ModuleClass component = backend.create(ModuleClass.class);
+    assertThat(component.string()).isEqualTo("foo");
+  }
+
+  @Test public void moduleClassAndInterfaceHierarchy() {
+    ModuleClassAndInterfaceHierarchy component =
+        backend.create(ModuleClassAndInterfaceHierarchy.class);
+    assertThat(component.string()).isEqualTo("foo");
+  }
+
+  @Test public void moduleClassAndInterfaceDuplicatesHierarchy() {
+    ModuleClassAndInterfaceDuplicatesHierarchy component =
+        backend.create(ModuleClassAndInterfaceDuplicatesHierarchy.class);
+    assertThat(component.string()).isEqualTo("foo");
+  }
+
+  @Test public void moduleClassHierarchy() {
+    ModuleClassHierarchy component = backend.create(ModuleClassHierarchy.class);
+    assertThat(component.string()).isEqualTo("foo");
+  }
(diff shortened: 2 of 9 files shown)


reflect/src/main/java/dagger/reflect/ReflectiveModuleParser.java
```diff
@@ -72,7 +84,12 @@ static void parse(Class<?> moduleClass, @Nullable Object instance,
 
         graphBuilder.add(key, binding);
       }
-      target = target.getSuperclass();
+
+      Class<?> superclass = target.getSuperclass();
+      if (superclass != Object.class && superclass != null) {
```
TWiStErRob: add `Object.class` and `null` to initial value of `seen` and let it `continue;`?

reflect/src/main/java/dagger/reflect/ReflectiveModuleParser.java
```diff
@@ -72,7 +84,12 @@ static void parse(Class<?> moduleClass, @Nullable Object instance,
...
+        queue.add(superclass);
+      }
+      Collections.addAll(queue, target.getInterfaces());
```
TWiStErRob: would it worth testing multiple inheritance?
i.e. `abstract class Foo extends Bar implements Baz, Quz`

reflect/src/main/java/dagger/reflect/ReflectiveModuleParser.java
JakeWharton: I'd prefer to keep these out of the collections altogether since they're more like bad data

reflect/src/main/java/dagger/reflect/ReflectiveModuleParser.java
JakeWharton: Added a test for this by way of testing that when an interface is included twice via a base class that it still works

reflect/src/main/java/dagger/reflect/ReflectiveModuleParser.java
TWiStErRob: edge case: if the method is called with either of these, it can get inside the collections, and the error won't be graceful

integration-tests/src/main/java/com/example/ModuleClassHierarchy.java
```diff
@@ -0,0 +1,23 @@
...
+
+  @Module
+  abstract class Module1 extends BaseModule {
+    @Provides static String string() {
+      return "foo";
+    }
+  }
+
+  @Module
+  abstract class BaseModule {
```
TWiStErRob: I think the `@Module` annotation on the base classes is not necessary, because it's not used in `modules =`

reflect/src/main/java/dagger/reflect/ReflectiveModuleParser.java
TWiStErRob: I was going for more the multiple interface to make sure all of the interfaces are checked on a single class not just the first one, but up to you

integration-tests/src/main/java/com/example/ModuleClassHierarchy.java
JakeWharton: The compiler requires all types with providers to contain it

integration-tests/src/test/java/com/example/IntegrationTest.java
```diff
@@ -256,6 +256,43 @@
...
+  }
+
+  @Test public void moduleInterfaceHierarchy() {
+    ModuleInterface component = backend.create(ModuleInterface.class);
+    assertThat(component.string()).isEqualTo("foo");
+  }
+
   private void ignoreReflectionBackend() {
     assumeTrue("Not yet implemented for reflection backend", backend != Backend.REFLECT);
   }
```
TWiStErRob: the length of this file and package is getting scary, have you thought about splitting it to multiple tests and moving the input classes to packages? (e.g. this batch of new tests would go into `com.example.module_hierarchy` or similar)

reflect/src/main/java/dagger/reflect/ReflectiveModuleParser.java
JakeWharton: Sure but that's parameter validation which is a separate concern. If you pass null you get an NPE and if you pass `Object.class` you get an error that its methods don't have `@Provides`

integration-tests/src/test/java/com/example/IntegrationTest.java
JakeWharton: Not a priority to me, no. The methods are small and navigable enough with cmd+f12 for now

PR #17, opened 2019-02-04
Add tests for @Binds into multibinding which is not yet supported by reflect

JakeWharton:


JakeWharton:
All keys will work because of how their lookup works so we don't (and can't) exhaustively test them all. `unwrapValue` is missing coverage, yes. We're unlikely to need qualifier tests for binds specifically, but probably for multibinds in general.

PR #18, opened 2019-02-04
Add support for component dependencies

JakeWharton:
Didn't add tests for all the type hierarchy walking stuff

### DIFF
--- a/reflect/src/main/java/dagger/reflect/ReflectiveComponentParser.java
+++ b/reflect/src/main/java/dagger/reflect/ReflectiveComponentParser.java
@@ -1,0 +1,53 @@
+package dagger.reflect;
+
+import java.lang.annotation.Annotation;
+import java.lang.reflect.Method;
+import java.lang.reflect.Type;
+import java.util.ArrayDeque;
+import java.util.Collections;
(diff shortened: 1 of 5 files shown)


reflect/src/main/java/dagger/reflect/ReflectiveComponentParser.java
```diff
@@ -0,0 +1,53 @@
...
+import static dagger.reflect.Reflection.findQualifier;
+
+final class ReflectiveComponentParser {
+  private static final Binding<?>[] NO_BINDINGS = new Binding<?>[0];
+
+  static void parse(Class<?> moduleClass, Object instance,
+      BindingGraph.Builder graphBuilder) {
+    Set<Class<?>> seen = new LinkedHashSet<>();
+    Deque<Class<?>> queue = new ArrayDeque<>();
+    queue.add(moduleClass);
```
TWiStErRob: Should the hierarchy walking be extracted and shared between the two parsers in the future?

JakeWharton: Sure. Makes sense to de-duplicate once their both merged.

PR #22, opened 2019-02-05
Allow DaggerCodegen to create/build Components which are inner classes

TWiStErRob:


codegen/src/main/java/dagger/internal/DaggerCodegen.java
```diff
@@ -46,6 +47,25 @@
     }
   }
 
+  private static <C> String deduceImplementationClassName(Class<C> componentClass) {
```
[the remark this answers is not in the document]
JakeWharton: Inserts are actually terrible because they always cause a re-allocation and copy.

integration-tests/src/test/java/com/example/IntegrationTest.java
```diff
@@ -285,6 +285,30 @@
...
+    assertThat(component.string()).isEqualTo("foo");
+  }
+
+  @Test public void nestedComponentBuilder() {
+    NestedComponent.TheComponent component =
+        backend.builder(NestedComponent.TheComponent.Builder.class).build();
+    assertThat(component.string()).isEqualTo("foo");
+  }
+
+  @Test public void deeplyNestedComponent() {
```
JakeWharton: Do we need both of these? Let's just keep the deep one.

TWiStErRob: K, Will revise tonight

TWiStErRob:
Rebased on top of develop, squashed all into one commit, removed duplicate test as asked.

codegen/src/main/java/dagger/internal/DaggerCodegen.java
```diff
@@ -46,6 +47,25 @@
...
+    Deque<Class<?>> classes = new ArrayDeque<>();
```
TWiStErRob: Note ArrayDeque now, not LinkedList, because errorprone

PR #23, opened 2019-02-05
Deduplicate class hierarchy traversal code

JakeWharton:


reflect/src/main/java/dagger/reflect/Reflection.java
```diff
@@ -98,6 +103,30 @@ static Object tryInvoke(@Nullable Object instance, Method method, Object... argu
     }
   }
 
+  static void walkHierarchy(Class<?> root, ClassVisitor visitor) {
```
TWiStErRob: Have you considered not doing the `visitor`, and just returning `seen` after the `while` loop terminates? and letting the "users" loop through?

JakeWharton: Sure. I'll update to that.

PR #24, opened 2019-02-05
Add error-prone and NullAway

JakeWharton:
NullAway is still only set to warn, for now…

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,3 +1,5 @@
+import net.ltgt.gradle.errorprone.CheckSeverity
+
 buildscript {
   ext.versions = [
       'dagger': '2.21',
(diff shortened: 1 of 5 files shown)


build.gradle
```diff
@@ -39,5 +52,18 @@ allprojects {
...
+    errorproneJavac 'com.google.errorprone:javac:9+181-r4173-1'
+    errorprone 'com.google.errorprone:error_prone_core:2.3.2'
+    errorprone 'com.uber.nullaway:nullaway:0.6.4'
+  }
+
+  tasks.withType(JavaCompile).configureEach {
+    options.errorprone {
+      disableWarningsInGeneratedCode = true
+      check("JavaxInjectOnAbstractMethod", CheckSeverity.OFF)
+      option("NullAway:AnnotatedPackages", "dagger.reflect")
```
TWiStErRob: reflect/src/main/java/dagger/Dagger.java
codegen/src/main/java/dagger/Dagger.java
codegen/src/main/java/dagger/internal/DaggerCodegen.java

build.gradle
```diff
@@ -24,12 +26,23 @@ buildscript {
...
       'guava': 'com.google.guava:guava:27.0.1-jre',
   ]
+
+  repositories {
+    mavenCentral()
+    gradlePluginPortal()
+  }
+
+  dependencies {
+    classpath 'net.ltgt.gradle:gradle-errorprone-plugin:0.7'
```
TWiStErRob: no `ext.deps` on the new dependencies because they're only used once?

build.gradle
JakeWharton: Those packages aren't owned by us so I don't want to include them and give a false sense of security

build.gradle
JakeWharton: yes

PR #21, opened 2019-02-04
Add @BindsOptionalOf support

burntcookie90:

PR #26, opened 2019-02-06
Model optional keys during binding linking

JakeWharton:
This removes the need to special case optional bindings inside the graph. Instead, the unlinked optional binding can request its dependency key as being optional. Optional keys which are not found will be null in the linked binding array.

Mostly renames in the code. The only interesting bit is in Binding and BindingGraph.

### DIFF
--- a/reflect/src/main/java/dagger/reflect/Binding.java
+++ b/reflect/src/main/java/dagger/reflect/Binding.java
@@ -31,35 +31,29 @@ import static dagger.reflect.Reflection.findQualifier;
 import static dagger.reflect.Reflection.tryInstantiate;
 import static dagger.reflect.Reflection.tryInvoke;
 
-interface Binding<T> extends Provider<T> {
-  Key[] NO_DEPENDENCIES = new Key[0];
+interface Binding {
+  abstract class UnlinkedBinding implements Binding {
+    abstract LinkRequest request();
+    abstract LinkedBinding<?> link(LinkedBinding<?>[] dependencies);
+  }
 
-  boolean isLinked();
-  Key[] dependencies();
-  Binding<T> link(Binding<?>[] dependencies);
+  final class LinkRequest {
+    static final LinkRequest EMPTY = new LinkRequest(new Key[0]);
 
-  abstract class LinkedBinding<T> implements Binding<T> {
-    @Override public final boolean isLinked() {
-      return true;
+    final Key[] keys;
+    final boolean[] optionals;
+
+    LinkRequest(Key[] keys) {
+      this(keys, new boolean[keys.length]);
     }
 
-    @Override public final Key[] dependencies() {
-      throw new UnsupportedOperationException();
-    }
-
-    @Override public final Binding<T> link(Binding<?>[] dependencies) {
-      throw new UnsupportedOperationException();
+    LinkRequest(Key[] keys, boolean[] optionals) {
+      this.keys = keys;
+      this.optionals = optionals;
     }
   }
 
-  abstract class UnlinkedBinding<T> implements Binding<T> {
-    @Override public final boolean isLinked() {
-      return false;
-    }
-
-    @Override public final T get() {
-      throw new UnsupportedOperationException();
-    }
+  abstract class LinkedBinding<T> implements Binding, Provider<T> {
   }
 
   final class Instance<T> extends LinkedBinding<T> {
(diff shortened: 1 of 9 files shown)


reflect/src/main/java/dagger/reflect/BindingGraph.java
```diff
@@ -15,105 +15,120 @@
...
         }
-
-        dependency = performLinking(dependencyKey, dependency, chain);
+        linkedBinding = performLinking(requestKey, (UnlinkedBinding) dependency, chain);
+      } else if (isOptional) {
+        linkedBinding = null;
+      } else {
+        StringBuilder builder = new StringBuilder("Cannot locate binding for ");
+        builder.append(requestKey).append('\n');
+        for (Map.Entry<Key, Binding> entry : chain.entrySet()) {
```
TWiStErRob: extract this to a method and reuse above as well? something like:
here:
`appendRequests(builder, chain, requestKey, "was not found");`
and above:
`appendRequests(builder, chain, requestKey, "forms a cycle");`

it would reduce complexity and dedupe, + the method would fit on one 1080p screen :)

reflect/src/main/java/dagger/reflect/ReflectiveMembersInjector.java
```diff
@@ -38,8 +39,7 @@
     Deque<ClassBindings> hierarchyBindings = new ArrayDeque<>();
```
TWiStErRob:
```suggestion
     Deque<ClassProviders> hierarchyProviders = new ArrayDeque<>();
```
to be consistent with field|methodProviders?

reflect/src/main/java/dagger/reflect/Binding.java
```diff
@@ -150,36 +142,36 @@ public LinkedOptionalBinding(Binding<?>[] dependencies) {
...
       Key[] dependencies = new Key[parameterTypes.length];
       for (int i = 0; i < parameterTypes.length; i++) {
         dependencies[i] = Key.of(findQualifier(parameterAnnotations[i]), parameterTypes[i]);
       }
-      return dependencies;
+      return new LinkRequest(dependencies);
     }
 
-    @Override public Binding<T> link(Binding<?>[] dependencies) {
+    @Override public LinkedBinding link(LinkedBinding<?>[] dependencies) {
```
TWiStErRob: rawtypes
```suggestion
    @Override public LinkedBinding<?> link(LinkedBinding<?>[] dependencies) {
```

reflect/src/main/java/dagger/reflect/Binding.java
```diff
@@ -115,33 +109,31 @@
...
     @Override
     public Optional<T> get() {
-      Binding<?> dependency = dependencies[0];
-      if (dependency == null) {
-        return Optional.empty();
-      }
-      return Optional.of((T) dependency.get());
+      return dependency == null
+          ? Optional.empty()
+          : Optional.of((T) dependency.get());
```
TWiStErRob: `dependency.get()` can be `null` (you were wondering here: https://github.com/JakeWharton/dagger-reflect/pull/21/commits/e3d9abd255edfc639110e483113d747775cfce34#r253727983):
```
java.lang.NullPointerException
	at java.util.Objects.requireNonNull(Objects.java:203)
	at java.util.Optional.<init>(Optional.java:96)
	at java.util.Optional.of(Optional.java:108)
	at dagger.reflect.Binding$LinkedOptionalBinding.get(Binding.java:132)
```
codegen throws: `java.lang.NullPointerException: Cannot return null from a non-@Nullable @Provides method`
to repro change `OptionalBinding` test to:
```java
@Provides static String foo() { return sneaky(); }
static String sneaky() { return null; }
```

reflect/src/main/java/dagger/reflect/ReflectiveModuleParser.java
```diff
@@ -42,7 +42,7 @@ static void parse(Class<?> moduleClass, @Nullable Object instance,
           throw new IllegalArgumentException("Private module methods are not allowed: " + method);
         }
 
-        Binding<?> binding;
+        Binding binding;
```
TWiStErRob:
```suggestion
        UnlinkedBinding binding;
```
note: needs import

JakeWharton: Inserting something that's already linked would be valid. There may not be any now, but there could be in the future.

reflect/src/main/java/dagger/reflect/ReflectiveMembersInjector.java
JakeWharton: This object isn't queried so I don't think naming it like a collection of something makes sense. It's a binder, maybe. I don't know. Can be figured out in a follow-up.

reflect/src/main/java/dagger/reflect/Binding.java
JakeWharton: Yes but it's existing behavior. Should be fixed separately.

reflect/src/main/java/dagger/reflect/BindingGraph.java
JakeWharton: I'm probably going to extract both to individual static methods for creating the exception so it can be re-used for `getProvider`. For a follow-up, though.

PR #29, opened 2019-02-07
Add test coverage for codegen frontend

JakeWharton:


JakeWharton:
I don't follow. None of these tests deal with visibility.

TWiStErRob:
DaggerCodegen reflects on some classes that are in the same package and hence visible and deemed accessible by the JVM. Reflecting on classes that are in other packages and package visible could lead to IllegalAccessError as far as I know, so by just moving the tests to `com.example` this would be caught, if ever happens. Having tests in separate package would resemble the real world more where users would be `create`ing their own Components.

JakeWharton:
Sure but that's not tied to this PR which is for adding coverage for exception cases.

PR #31, opened 2019-02-07
Fix typo in test, use the right class to test

TWiStErRob:


integration-tests/src/test/java/com/example/IntegrationTest.java
```diff
@@ -388,7 +388,7 @@
   }
 
   @Test public void moduleInterfaceHierarchy() {
```
TWiStErRob: was identical to the method above: `moduleInterface`

PR #33, opened 2019-02-08
Move binding types to top-level

JakeWharton:
This separates the implementations from abstract types in preparation for refactoring and multibinds


reflect/src/main/java/dagger/reflect/UnlinkedJustInTimeBinding.java
```diff
@@ -0,0 +1,52 @@
...
+import java.lang.reflect.Constructor;
+import java.lang.reflect.Field;
+import java.lang.reflect.Method;
+import java.lang.reflect.Type;
+import javax.inject.Inject;
+
+import static dagger.reflect.DaggerReflect.notImplemented;
+import static dagger.reflect.Reflection.findQualifier;
+
+final class UnlinkedJustInTimeBinding<T> extends UnlinkedBinding {
```
TWiStErRob: Class was referenced in comment:
https://github.com/JakeWharton/dagger-reflect/blob/09bb185fa8c478a6c97c751d9157013ebc6f7718/reflect/src/main/java/dagger/reflect/ReflectiveMembersInjector.java#L126

`dagger.reflect.Binding.UnlinkedJustInTime.dependencies` is now `dagger.reflect.UnlinkedJustInTimeBinding.request`

JakeWharton: I removed the reference in #40 since it's likely to become stale again in the future. Plus that type will be used in places where constructor injection isn't happening anyway.

PR #34, opened 2019-02-08
Fix errorprone UndefinedEquals violations

TWiStErRob:
```
@SuppressWarnings("UndefinedEquals") // comparing `@Binds CharSequence provided(String)` to `String`
```
could be an alternative solution, but then other newer violations may be hidden and actually give a false-positive pass

The current PR approach of adding a cast is good both ways:
 * if it's unnecessary IDEA's `RedundantCast` kicks in
 * if cast is not there `UndefinedEquals` shows up again

### DIFF
--- a/integration-tests/src/test/java/com/example/IntegrationTest.java
+++ b/integration-tests/src/test/java/com/example/IntegrationTest.java
@@ -45,7 +45,7 @@ public final class IntegrationTest {
 
   @Test public void bindsProvider() {
     BindsProvider component = backend.create(BindsProvider.class);
-    assertThat(component.string()).isEqualTo("foo");
+    assertThat(component.number()).isEqualTo(42);
   }
 
   @Test public void bindsProviderNull() {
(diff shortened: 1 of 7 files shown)


JakeWharton:
Maybe we use different types instead? We could bind to `Object` instead?

TWiStErRob:
Good idea; wanted to keep it realistic, so used two types that extend, but not Object.

Also found a missing feature in the process, will open PR with failing tests.

integration-tests/src/test/java/com/example/IntegrationTest.java
```diff
@@ -40,7 +40,7 @@
 
   @Test public void bindsProvider() {
     BindsProvider component = backend.create(BindsProvider.class);
-    assertThat(component.string()).isEqualTo("foo");
+    assertThat((String) component.string()).isEqualTo("foo");
```
JakeWharton: What about this one?

TWiStErRob: huh... fixed

PR #35, opened 2019-02-08
Clean up build warnings

TWiStErRob:
... and prevent future ones.

Didn't want to go straight for adding `-Werror`, because it may clobber local hacking; but I think on CI the commands should include `-Pstrict.compilation=true`.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -120,6 +120,24 @@ subprojects {
         assertsEnabled = true
       }
     }
+
+    options.compilerArgs += [
+        // enable all warnings
+        '-Xlint:all',
+        // hide warning: No processor claimed any of these annotations: dagger.*,org.junit.Test
+        '-Xlint:-processing',
+        // hide warning: MethodParameters attribute introduced in version 52.0 class files is ignored in version 51.0 class files
+        '-Xlint:-classfile',
+        // hide warning: auxiliary class C in full/path/to/C.java should not be accessed from outside its own source file
+        '-Xlint:-auxiliaryclass'
+    ]
+
+    if (project.findProperty("strict.compilation") == "true") {
+      options.compilerArgs += [
+          // strict compilation, treat all warnings as errors, including errorprone
+          '-Werror'
+      ]
+    }
   }
 
   plugins.withId('java-library') {
--- a/build.gradle
+++ b/build.gradle
@@ -101,6 +101,8 @@ subprojects {
   }
 
   tasks.withType(JavaCompile).configureEach {
+    options.encoding = 'UTF-8'
+
     options.errorprone {
       disableWarningsInGeneratedCode = true
       excludedPaths = ".*/build/generated/.*"
--- a/integration-tests/upstream/build.gradle
+++ b/integration-tests/upstream/build.gradle
@@ -106,3 +106,7 @@ def ensureDaggerSubmodule = tasks.create('ensureDaggerSubmodule') {
   }
 }
 compileTestJava.dependsOn(ensureDaggerSubmodule)
+
+tasks.withType(JavaCompile).configureEach {
+  options.encoding = 'UTF-8'
+}
--- a/reflect/src/main/java/dagger/reflect/UnlinkedMapOfProviderBinding.java
+++ b/reflect/src/main/java/dagger/reflect/UnlinkedMapOfProviderBinding.java
@@ -13,11 +13,10 @@ final class UnlinkedMapOfProviderBinding extends UnlinkedBinding {
   }
 
   @Override
-  public LinkedBinding<Map<Object, Provider<Object>>> link(Linker linker, Scope scope) {
-    Map<Object, Provider<Object>> mapOfProviders = new LinkedHashMap<>(entryBindings.size());
+  public LinkedBinding<?> link(Linker linker, Scope scope) {
+    Map<Object, Provider<?>> mapOfProviders = new LinkedHashMap<>(entryBindings.size());
     for (Map.Entry<Object, Binding> entryBinding : entryBindings.entrySet()) {
-      LinkedBinding<Object> binding =
-          (LinkedBinding<Object>) entryBinding.getValue().link(linker, scope);
+      LinkedBinding<?> binding = entryBinding.getValue().link(linker, scope);
       mapOfProviders.put(entryBinding.getKey(), binding);
     }
     return new LinkedInstanceBinding<>(mapOfProviders);
--- a/integration-tests/android/build.gradle
+++ b/integration-tests/android/build.gradle
@@ -11,6 +11,7 @@ android {
 
     versionCode 1
     versionName '1'
+    javaCompileOptions.annotationProcessorOptions.includeCompileClasspath = false
   }
 
   compileOptions {
--- a/integration-tests/android/src/main/AndroidManifest.xml
+++ b/integration-tests/android/src/main/AndroidManifest.xml
@@ -1,11 +1,15 @@
 <manifest
     xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:tools="http://schemas.android.com/tools"
     package="com.example"
     >
 
   <application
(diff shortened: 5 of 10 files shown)


reflect/src/main/java/dagger/reflect/ComponentInvocationHandler.java
```diff
@@ -75,6 +75,7 @@ private ComponentInvocationHandler(BindingGraph graph) {
             "Members injection methods may only return the injected type or void: " + method);
       }
 
+      @SuppressWarnings({"unchecked", "RedundantCast"})
```
TWiStErRob: see https://youtrack.jetbrains.com/issue/IDEA-206560

JakeWharton: can you leave this as a comment in source?

TWiStErRob: done

also fixed 2 dangling alignment issues like this:
https://github.com/JakeWharton/dagger-reflect/compare/9230378380e3d07a5ecbb3b3dad3d77b1e24b798..bf291c4220b14819da46e81f9e44c9fe9e9d823c#diff-6f0b84cc3cf4faa16145ef4af80a80daL41

JakeWharton:
I've merged e62ad1ba6ef3c6dc52350c95dab10cba84a367f8 and 4f75adaf65d222dabe3c010ee7566dca7f3994be. I'm still thinking about the others. I think bf2 might be obsoleted by a change I'm about to make to bindings and I need to think about 0c4 more. I'm all for more of the error-prone warnings but I'm a bit skeptical of the java ones. Maybe we can turn it on until they annoy me.

TWiStErRob:
0c4e911 is unlikely to produce anything other than unchecked, rawtypes and deprecation, which all would be visible in IDEA anyway as warnings, so quick to notice and usually very easy to fix. If "unlikely" is not good enough, it's always an option to flip it around from `-Xlint:all -Xlint:-processing -Xlint:-classfile` to `-Xlint:unchecked -Xlint:rawtypes -Xlint:deprecation`.

TWiStErRob:
Rebased and fixed all of the current warnings. `gradlew clean check` runs without warnings, except:

This unfixable warning from AGP:
```
> Task :integration-tests:android:processCodegenDebugResources
Changing the value for a property with a final value has been deprecated. This will fail with an error in Gradle 6.0.
```

and this Errorprone warning which is fixed in #156
```
> Task :reflect:compileJava
P:\projects\contrib\github-dagger2-reflect\reflect\src\main\java\dagger\reflect\Scope.java:82: warning: [ObjectToString] Linker is final and does not override Object.toString, so converting it to a string will print its identity (e.g. `Linker@ 4488aabb`) instead of useful information
.
                + linker); // TODO nice error message with scope chain
                  ^
    (see https://errorprone.info/bugpattern/ObjectToString)
1 warning
```

build.gradle
```diff
@@ -118,6 +120,24 @@ subprojects {
...
+        '-Xlint:all',
+        // hide warning: No processor claimed any of these annotations: dagger.*,org.junit.Test
+        '-Xlint:-processing',
+        // hide warning: MethodParameters attribute introduced in version 52.0 class files is ignored in version 51.0 class files
+        '-Xlint:-classfile',
+        // hide warning: auxiliary class C in full/path/to/C.java should not be accessed from outside its own source file
+        '-Xlint:-auxiliaryclass'
+    ]
+
+    if (project.property("strict.compilation").toBoolean()) {
```
JakeWharton: Can we do something like `findProperty("string.compilation") == "true"` to avoid having to change `gradle.properties`?

JakeWharton (commented):
Mostly looks good. Just some nits.

integration-tests/upstream/build.gradle
```diff
@@ -106,3 +106,7 @@ def ensureDaggerSubmodule = tasks.create('ensureDaggerSubmodule') {
   }
 }
 compileTestJava.dependsOn(ensureDaggerSubmodule)
+
+tasks.withType(JavaCompile).configureEach {
+  options.encoding = 'UTF-8'
```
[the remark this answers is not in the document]
TWiStErRob: No, it's not, because of
```
  if (project == project(':integration-tests:upstream')) {
    return
  }
```

build.gradle
TWiStErRob: Sure, if you prefer that. The reason behind a default value in gradle.properties was to be explicit; cleanly tell: "this key exists" and "this is the default value". This way multiple usages of the same property can be defaulted by using Gradle's features, instead of potentially duplicating code, just used to doing this nowadays.

reflect/src/main/java/dagger/reflect/UnlinkedMapOfLazyBinding.java
```diff
@@ -13,6 +13,8 @@
 
   @Override
   public LinkedBinding<?> link(Linker linker, Scope scope) {
+    // assume that mapOfProviderKey is Map<K, Provider<V>> and linker returns the correct Binding
```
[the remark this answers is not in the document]
TWiStErRob: Beautiful. Thank you for the link!

TWiStErRob:
Double-force-pushed to resolve the conflict (deleted file in another PR) and see the code review changes:
https://github.com/JakeWharton/dagger-reflect/compare/a524045a02c38ec895911264f989ef7d7ea9e3fa..1efaa5293e48123358781e6067cebf83574c3e91

reflect/src/main/java/dagger/reflect/UnlinkedMapOfProviderBinding.java
```diff
@@ -13,11 +13,10 @@
   }
 
   @Override
-  public LinkedBinding<Map<Object, Provider<Object>>> link(Linker linker, Scope scope) {
```
JakeWharton: This is unfortunate. Not that I expect anyone to ever see these types, and they're erased anyway, but it at least forces a bit of safety into the method. Doing `LinkedBinding<Map<?, Provider<?>>>` doesn't work?

build.gradle
```diff
@@ -118,6 +120,24 @@ subprojects {
...
     }
+
+    options.compilerArgs += [
+        // enable all warnings
+        '-Xlint:all',
+        // hide warning: No processor claimed any of these annotations: dagger.*,org.junit.Test
+        '-Xlint:-processing',
+        // hide warning: MethodParameters attribute introduced in version 52.0 class files is ignored in version 51.0 class files
+        '-Xlint:-classfile',
+        // hide warning: auxiliary class C in full/path/to/C.java should not be accessed from outside its own source file
```
JakeWharton: What triggers all of these? I'm less interested in the description since that can be Googled but more interested in what causes them and why it's safe to ignore. Similar to an `@Ignore` for a test. The 'processing' one has a good justification but not these last two.

JakeWharton:
I've merged and cherry-picked about 7 or 8 of the commits in this PR.

TWiStErRob:
> I've merged and cherry-picked about 7 or 8 of the commits in this PR.

ok, thanks; will rebase and continue to discuss the rest

build.gradle
TWiStErRob: All 3 are the warning messages, not explanation. I'll research and explain them in the context of this project.

TWiStErRob:
Only two relevant commits remain:
 * https://github.com/JakeWharton/dagger-reflect/pull/35/commits/472480b037e07d3c31f552d96d9ff5dee2c25ef0 which would be probably solved by a Gradle/AGP upgrade
 * https://github.com/JakeWharton/dagger-reflect/pull/35/commits/1efaa5293e48123358781e6067cebf83574c3e91 which needs reword based on https://github.com/JakeWharton/dagger-reflect/pull/35#discussion_r310857269

Closing this until the project becomes active again.

PR #36, opened 2019-02-08
Add tests for @Provides @Nullable and @Binds @Nullable

TWiStErRob:


integration-tests/src/main/java/com/example/BindsProviderNull.java
```diff
@@ -0,0 +1,21 @@
...
+interface BindsProviderNull {
+
+  @Nullable CharSequence string();
+
+  @Module
+  abstract class Module1 {
+    @Provides static @Nullable String string() {
+      return null;
+    }
+    @Binds abstract CharSequence charSequence(@Nullable String foo);
```
JakeWharton: Interesting... Seems like Dagger should require `@Nullable` here.

PR #37, opened 2019-02-08
Fix some inspections

TWiStErRob:


JakeWharton:
I have merged 3d077c10319bf1e4d132aa4f677587a0ad9e6797 and dd7dfc104bbb1019e39b877d104cd8f6475affa6.

I think type widening only makes sense for public API. In this case, it only makes sense to search for component types inside another type and module and dependency classes should be unique.

PR #38, opened 2019-02-08
Enable parallel build

TWiStErRob:
Measurement: `gradlew clean check`
Current `master`: 16s (https://scans.gradle.com/s/im3x27mkieat6)
Gradle parallel: 10s (https://scans.gradle.com/s/xbixn2tzo6va4)

I like playing with builds, so I tried to go further:
Move APT classes to modules: 7s (https://scans.gradle.com/s/vutaiobangdts)
Move IntegrationTest next to classes: 6s (https://scans.gradle.com/s/5wiwkyulfkuy2)

The last two steps are not included in this PR, but the commits are here:
https://github.com/TWiStErRob/dagger-reflect/commits/abandoned/parallel
if you think it would be useful, happy to clean it up and raise a PR

Mind you I have a brand new PC with 14 cores, dagger-reflect builds use 4-8 cores only, so YMMV significantly.

Also tried https://guides.gradle.org/performance/#parallel_test_execution, but makes no difference with current amount of tests, compilation is the slowest.

Also tried `-Adagger.formatGeneratedSource=disabled` but makes no difference

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,3 +1,4 @@
+# Build Properties
 GROUP=com.jakewharton.dagger
 VERSION_NAME=0.1.0-SNAPSHOT
 
@@ -14,3 +15,6 @@ POM_LICENCE_DIST=repo
 
 POM_DEVELOPER_ID=jakewharton
 POM_DEVELOPER_NAME=Jake Wharton
+
+# Build Configuration
+org.gradle.parallel=true


JakeWharton:
So we can't check this in because it allows the `uploadArchives` task to run in parallel. Sonatype, the target for the artifacts, sees concurrent uploads as separate and creates two separate staging repositories. This will break the snapshot and release builds as a result. Those tasks need restricted to run serially before something like this can be merged.

TWiStErRob:
Do you mean something like this?
```kotlin
val uploads = allprojects.mapNotNull { it.tasks.uploadArchives }
uploads.zip(uploads.skip(1)) { task1, task2 -> task1.mustRunAfter(task2) }
```
but Groovy and hopefully using lazy task getters :)
Assuming order of them don't matter, just that they're serial.

JakeWharton:
That might work. There's some other options from the feature request on Gradle: https://github.com/gradle/gradle/issues/7047#issuecomment-430139316

JakeWharton:
Speed is not a concern on CI

TWiStErRob:
Agree, the point here was to utilize Gradle features to speed up local development as much as possible. I guess I can try some global override from my home directory, since the uploadArchives is not an issue locally.

PR #40, opened 2019-02-09
Rename ClassBindings to ClassInjector to reflect its behavior

JakeWharton:


reflect/src/main/java/dagger/reflect/ReflectiveMembersInjector.java
```diff
@@ -87,43 +87,43 @@
...
         Map<Field, Provider<?>> fieldProviders,
         Map<Method, Provider<?>[]> methodProviders) {
       this.fieldProviders = fieldProviders;
       this.methodProviders = methodProviders;
     }
 
-    void injectMembers(Object instance) {
-      // [@Inject] Constructors are injected first, followed by fields, and then methods.
-      // Note: the constructor injection is in dagger.reflect.Binding.UnlinkedJustInTime.dependencies
+    @Override public void injectMembers(T instance) {
```
TWiStErRob: ah, that's why it was `T`, the interface

reflect/src/main/java/dagger/reflect/ReflectiveMembersInjector.java
```diff
@@ -87,43 +87,43 @@
...
-    return new ReflectiveMembersInjector<>(hierarchyBindings);
+    return new ReflectiveMembersInjector<>(classInjectors);
   }
 
-  private final Iterable<ClassBindings> classBindings;
+  private final Iterable<ClassInjector<T>> classBindings;
 
-  private ReflectiveMembersInjector(Iterable<ClassBindings> classBindings) {
+  private ReflectiveMembersInjector(Iterable<ClassInjector<T>> classBindings) {
     this.classBindings = classBindings;
```
TWiStErRob: field and ctor param is still "bindings"

reflect/src/main/java/dagger/reflect/ReflectiveMembersInjector.java
```diff
@@ -87,43 +87,43 @@
       }
 
       if (!fieldProviders.isEmpty() || !methodProviders.isEmpty()) {
-        // [@Inject] Fields and methods in superclasses are injected before those in subclasses.
-        // we are walking up (getSuperclass), but spec says @Inject should be walking down the hierarchy
-        hierarchyBindings.addFirst(new ClassBindings(fieldProviders, methodProviders));
+        // Per JSR 330, fields and methods in superclasses are injected before those in subclasses.
+        // Since we are traversing upward in the hierarchy each injector is thus prepended to the
```
TWiStErRob: "each injector is" doesn't make sense to me. There's something missing around there, it may be just commas.

reflect/src/main/java/dagger/reflect/ReflectiveMembersInjector.java
JakeWharton: Yeah there's no real reason to do this, but it was effectively free so why not

reflect/src/main/java/dagger/reflect/ReflectiveMembersInjector.java
```diff
@@ -87,43 +87,43 @@
...
         Map<Method, Provider<?>[]> methodProviders) {
       this.fieldProviders = fieldProviders;
       this.methodProviders = methodProviders;
     }
 
-    void injectMembers(Object instance) {
-      // [@Inject] Constructors are injected first, followed by fields, and then methods.
-      // Note: the constructor injection is in dagger.reflect.Binding.UnlinkedJustInTime.dependencies
+    @Override public void injectMembers(T instance) {
+      // Per JSR 330, fields are injected before methods.
```
burntcookie90: Where did you see this? `Inject` docs say 
> Ordering of injection among fields and among methods in the same class is not specified.

TWiStErRob: https://docs.oracle.com/javaee/6/api/javax/inject/Inject.html

> among fields and among methods in the same class

this means that the `fieldProviders` and `methodProviders` iteration order can be unspecified, which it is, because `getDeclaredFields` and `getDeclaredMethods` doesn't guarantee any ordering.

this comment refers to the previous sentence in Inject javadoc:
> Constructors are injected first, followed by fields, and then methods. Fields and methods in superclasses are injected before those in subclasses.

and it specifies that `fieldProviders` must be iterated before `methodProviders`

PR #41, opened 2019-02-12
Support JIT binding of parameterized types

JakeWharton:


reflect/src/main/java/dagger/reflect/ReflectiveJustInTimeProvider.java
```diff
@@ -18,10 +19,14 @@
...
 
     Type type = key.type();
-    if (!(type instanceof Class<?>)) {
-      return null; // Parameterized/array types can't be just-in-time satisfied.
+    Class<Object> cls;
+    if (type instanceof ParameterizedType) {
+      cls = (Class<Object>) ((ParameterizedType) type).getRawType();
+    } else if (type instanceof Class<?>) {
+      cls = (Class<Object>) type;
+    } else {
```
TWiStErRob: based on `dagger.reflect.Types#getRawType` there could be other cases: this `else` covers not just arrays, but other things (not exactly sure what). Have you considered this?:
```
else if (isArray(type)) return null
else notImplemented/IllegalArgument
```

TWiStErRob: Checking deeper, I guess `BindingGraph.findBinding` will return null and an exception will be thrown there?

JakeWharton: For just-in-time the only other one that makes sense is probably a wildcard, although from what I remember Dagger doesn't do well with wildcards. Not sure how Dagger would handle it handle.

And yes throwing a proper exception is not the responsibility of this code because it doesn't know about the request chain to get here. It just knows whether it can load or not.

TWiStErRob: changing to `Thing<? extends String> thing();`:
```
error: [Dagger/MissingBinding]
com.example.JustInTimeGeneric.Thing<? extends java.lang.String>
cannot be provided without an @Provides-annotated method.
```
so I guess it doesn't support it for JIT, but if there's was a provider with the right key, it would work.

JakeWharton: Yep. And that _should_ already work with a normal provider.

PR #42, opened 2019-02-12
Let unlinked bindings resolve dependencies directly while linking

JakeWharton:
This change will make linking future multibindings simpler.

As part of this change, tease apart the concept of the key->binding map and a "scope". In the future, Scope (and a Scope.Builder) might be public types for creating graphs from things other than reflection.


JakeWharton:
I'll write actual docs on these types... eventually...

JakeWharton:
Multibindings were really hard in the old architecture. With this it's _greatly_ simplified (coming in a future PR).

reflect/src/main/java/dagger/reflect/BindingMap.java
```diff
@@ -0,0 +1,95 @@
...
+    }
+    return linked;
+  }
+
+  static final class Builder {
+    private final Map<Key, Binding> bindings = new LinkedHashMap<>();
+    private JustInTimeProvider jitProvider = JustInTimeProvider.NONE;
+
+    Builder justInTimeProvider(JustInTimeProvider jitProvider) {
+      if (jitProvider == null) throw new NullPointerException("jitProvider == null");
```
burntcookie90: lol

PR #44, opened 2019-02-12
Add animal sniffer for reflect runtime

JakeWharton:


JakeWharton:
Not sure why this misses the `Optional.class` literal in `ReflectiveModuleParser` but better than nothing.

reflect/src/main/java/dagger/reflect/Key.java
```diff
@@ -31,9 +31,30 @@ static Key of(@Nullable Annotation qualifier, Type type) {
 
   @Override public final String toString() {
     Annotation qualifier = qualifier();
-    String type = type().getTypeName(); // TODO 1.8 method
+    String type = typeName();
```
TWiStErRob: `TypeUtil.canonicalize(this).toString()`?

JakeWharton: I don't like that this gives the idea that canonicalization is required when really we're just backporting a Java 8 method. I'll change to a static method with a comment to indicate it's just a backport and we should switch to `Type.getTypeName()` once available.

PR #47, opened 2019-02-14
Add support for @IntoSet bindings

JakeWharton:


reflect/src/main/java/dagger/reflect/BindingMap.java
```diff
@@ -80,9 +85,37 @@ Builder add(Key key, Binding binding) {
...
       ConcurrentHashMap<Key, Binding> allBindings = new ConcurrentHashMap<>(bindings);
+
+      // Coalesce all of the bindings for each key into a single set binding.
+      for (Map.Entry<Key, List<Binding>> setBinding : setBindings.entrySet()) {
+        Key elementKey = setBinding.getKey();
+        Key setKey = Key.of(elementKey.qualifier(),
+            new ParameterizedTypeImpl(null, Set.class, elementKey.type()));
+
+        // Take a defensive copy in case the builder is being re-used.
+        List<Binding> value = new ArrayList<>(setBinding.getValue());
```
JakeWharton: Whoops. This should probably get a better name. Probably the default from extract variable intention action.

PR #50, opened 2019-02-14
Add support for @IntoMap

JakeWharton:
Need a test for `unwrapValue=false` but that can come later.

### DIFF
--- a/reflect/src/main/java/dagger/reflect/BindingMap.java
+++ b/reflect/src/main/java/dagger/reflect/BindingMap.java
@@ -65,6 +65,7 @@ final class BindingMap {
   static final class Builder {
     private final Map<Key, Binding> bindings = new LinkedHashMap<>();
     private final Map<Key, List<Binding>> setBindings = new LinkedHashMap<>();
+    private final Map<Key, Map<Object, Binding>> mapBindings = new LinkedHashMap<>();
     private JustInTimeProvider jitProvider = JustInTimeProvider.NONE;
 
     Builder justInTimeProvider(JustInTimeProvider jitProvider) {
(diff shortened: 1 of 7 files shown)


reflect/src/main/java/dagger/reflect/BindingMap.java
```diff
@@ -98,6 +99,23 @@ Builder addIntoSet(Key key, Binding binding) {
       return this;
     }
 
+    Builder addIntoMap(Key key, Object mapKey, Binding mapValueBinding) {
```
JakeWharton: Maybe using 'entry' would be a good prefix. Similar to 'element' for sets.

burntcookie90: `bindingKey`?

reflect/src/main/java/dagger/reflect/ReflectiveModuleParser.java
```diff
@@ -69,11 +72,44 @@ static void parse(Class<?> moduleClass, @Nullable Object instance,
...
+      keyValue = Reflection.tryInvoke(keyAnnotation, keyAnnotationMethod);
+      if (keyValue == null) {
+        throw new AssertionError(); // Not allowed by the Java language specification.
+      }
+    } else {
+      keyType = keyAnnotationType;
+      keyValue = keyAnnotation;
+    }
+
+    Key key = Key.of(valueKey.qualifier(),
```
burntcookie90: key key key key key key

reflect/src/main/java/dagger/reflect/UnlinkedMapBinding.java
```diff
@@ -0,0 +1,27 @@
...
+    for (Map.Entry<Object, Binding> entryBinding : entryBindings.entrySet()) {
+      LinkedBinding<Object> linkedBinding =
+          (LinkedBinding<Object>) entryBinding.getValue().link(linker);
+      linkedBindings.put(entryBinding.getKey(), linkedBinding);
+    }
+    return new LinkedMapBinding<>(linkedBindings);
+  }
+
+  @Override public String toString() {
+    return "Set" + entryBindings;
```
burntcookie90: "Set" should be "Map"

reflect/src/main/java/dagger/reflect/BindingMap.java
JakeWharton: I could also just write some docs...

reflect/src/main/java/dagger/reflect/UnlinkedMapBinding.java
JakeWharton: Applied locally

reflect/src/main/java/dagger/reflect/BindingMap.java
JakeWharton: Settled on entryKey and entryValueBinding as least confusing... for now.

PR #58, opened 2019-02-18
Add support for simple subcomponent provision and scope hierarchy

JakeWharton:
Subcomponent builder provisions are supported and subcomponent provision provided it doesn't require any module instances.

### DIFF
--- a/reflect/src/main/java/dagger/reflect/ComponentInvocationHandler.java
+++ b/reflect/src/main/java/dagger/reflect/ComponentInvocationHandler.java
@@ -25,11 +25,14 @@ import java.util.concurrent.ConcurrentHashMap;
 import javax.inject.Provider;
 import org.jetbrains.annotations.Nullable;
 
-import static dagger.reflect.DaggerReflect.notImplemented;
 import static dagger.reflect.Reflection.findQualifier;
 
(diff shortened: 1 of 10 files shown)


reflect/src/main/java/dagger/reflect/ComponentInvocationHandler.java
```diff
@@ -86,9 +89,18 @@ private ComponentInvocationHandler(Scope scope) {
...
-          && ((Class<?>) returnType).getAnnotation(Subcomponent.class) != null) {
-        throw notImplemented("Subcomponents");
+      if (returnType instanceof Class<?>) {
+        Class<?> returnClass = (Class<?>) returnType;
+        if (returnClass.getAnnotation(Subcomponent.class) != null) {
+          return new SubcomponentMethodInvocationHandler(returnClass, scope);
+        }
+        // TODO if (returnClass.getAnnotation(Subcomponent.Builder.class) != null) {
+        // Until @Subcomponent.Builder has runtime retention, check whether this return type belongs
+        // to a subcomponent builder by looking for an enclosing class annotated with @Subcomponent.
```
JakeWharton: Already fixed internally at Google. Waiting for sync out to GitHub. Also fixed `@Component.Builder` and some other annotations so we can use them.

PR #59, opened 2019-02-19
Added an integration test for a final module

PaulWoitaschek:
Right now the reflection backend fails with 
`java.lang.IllegalStateException: com.example.ImplicitFinalModules.Module1 must be set`

However it should not as the module is no-arg and dagger parses it fine.

I ran into this scenario when testing dagger-reflect with my audiobook player and this is the commit that fixed it: https://github.com/PaulWoitaschek/Voice/commit/40e6daf20619470e599bcab29154bb45953adb81

### DIFF
--- a/integration-tests/src/test/java/com/example/IntegrationTest.java
+++ b/integration-tests/src/test/java/com/example/IntegrationTest.java
@@ -159,6 +159,14 @@ public final class IntegrationTest {
     assertThat(component.string()).isEqualTo("4");
   }
 
+  @Test public void implicitModules() {
+    ignoreReflectionBackend();
+
+    ImplicitFinalModules component = backend.create(ImplicitFinalModules.class);
+
+    assertThat(component.string()).isEqualTo(ImplicitFinalModules.Module1.VALUE);
+  }
+
   @Test public void builderExplicitModulesOmitted() {
     try {
       backend.builder(BuilderExplicitModules.Builder.class).build();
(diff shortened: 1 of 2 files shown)


integration-tests/src/test/java/com/example/IntegrationTest.java
```diff
@@ -1,15 +1,17 @@
...
 import org.junit.runner.RunWith;
 import org.junit.runners.Parameterized;
 import org.junit.runners.Parameterized.Parameter;
 import org.junit.runners.Parameterized.Parameters;
 
+import java.util.Map;
+import java.util.Optional;
+import java.util.Set;
+
+import javax.inject.Provider;
```
JakeWharton: Can you revert this order change?

integration-tests/src/test/java/com/example/IntegrationTest.java
```diff
@@ -159,6 +161,12 @@
     assertThat(component.string()).isEqualTo("4");
   }
 
+  @Test public void implicitModules() {
+    ImplicitFinalModules component = backend.create(ImplicitFinalModules.class);
```
JakeWharton: Add `ignoreReflectBackend()` as the first statement so we can land this test before the fix.

PaulWoitaschek: Done.

integration-tests/src/test/java/com/example/IntegrationTest.java
PaulWoitaschek: Done. Can you somehow ship the code style with the project so I can just press auto-format?

JakeWharton: It uses https://github.com/square/java-code-styles right now (as do all my projects). I've been debating about using google-java-format but it does some nasty things to builders. I should probably relent, though.

PR #61, opened 2019-02-22
Move JIT binding lookup to Scope

JakeWharton:
This is in prepartion for honoring scope annotations. If the JIT binding factory loads a class which is scoped we'll need to traverse the hierarchy to find which scope to add it to.


JakeWharton:
Also starting to add docs!

PR #64, opened 2019-02-28
Honor scope annotations on JIT bindings

JakeWharton:


reflect/src/main/java/dagger/reflect/Scope.java
```diff
@@ -62,6 +60,34 @@
...
 
+  /**
+   * Attempt to insert a binding for {@code key} as a result of a just-in-time lookup.
+   * <p>
+   * If {@code lookup} does not contain a scoping annotation or it contains a scoping annotation
+   * which matches the one for this scope, the binding will be inserted, linked, and returned.
+   * Otherwise, the parent scope (if any) will be recursively checked. If no matching scope
+   * annotation is found by traversing the parents null will be returned.
+   */
+  private @Nullable LinkedBinding<?> putJitBinding(Key key, JustInTimeLookup lookup) {
```
runningcode: where is this called from? :)

JakeWharton: Line 99

PR #69, opened 2019-03-07
Allow private helper methods in modules.

runningcode:
We have a couple of private helper methods in our modules.
This changes the behavior to allow private methods in modules if and
only if they have no annotations.

An alternative is to keep a blacklist of annotations which are not allowed
on private methods such as `@Provides`?

### DIFF
--- a/reflect/src/main/java/dagger/reflect/ReflectiveModuleParser.java
+++ b/reflect/src/main/java/dagger/reflect/ReflectiveModuleParser.java
@@ -29,12 +29,12 @@ final class ReflectiveModuleParser {
   static void parse(Class<?> moduleClass, @Nullable Object instance, Scope.Builder scopeBuilder) {
     for (Class<?> target : Reflection.getDistinctTypeHierarchy(moduleClass)) {
       for (Method method : target.getDeclaredMethods()) {
-        if ((method.getModifiers() & PRIVATE) != 0) {
-          throw new IllegalArgumentException("Private module methods are not allowed: " + method);
+        Annotation[] annotations = method.getAnnotations();
+        if ((method.getModifiers() & PRIVATE) != 0 && (annotations == null)) {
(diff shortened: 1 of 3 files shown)


reflect/src/main/java/dagger/reflect/ReflectiveModuleParser.java
```diff
@@ -29,12 +29,12 @@
   static void parse(Class<?> moduleClass, @Nullable Object instance, Scope.Builder scopeBuilder) {
     for (Class<?> target : Reflection.getDistinctTypeHierarchy(moduleClass)) {
       for (Method method : target.getDeclaredMethods()) {
-        if ((method.getModifiers() & PRIVATE) != 0) {
-          throw new IllegalArgumentException("Private module methods are not allowed: " + method);
+        Annotation[] annotations = method.getAnnotations();
+        if ((method.getModifiers() & PRIVATE) != 0 && (annotations == null)) {
```
JakeWharton: Hmm I'm not sure this check is suitable since if it were written in Kotlin you'd presumably get a `@NonNull` annotation. Ideally we only check when calling `addBinding` which is where we absolutely know you're trying to add a Dagger method.

Or, optionally, maybe we just don't validate this and let the real compiler catch it?

runningcode: Let me take a swing at moving the check in to `addBinding`.

runningcode: Since @Binds and @BindsOptionalOf must be abstract methods anyways, they can't be private, so I just added the check right before addBinds for `@Provides` annotated methods.

JakeWharton: Works for me!

PR #72, opened 2019-03-07
Change LinkedBinding to not be a Provider

JakeWharton:
We are going to start handling requests for `Provider<T>` soon and this will be confusing. A `LinkedBinding<Provider<T>>` would be a `Provider<Provider<T>>` then and I don't want to get caught up in accidentally interpreting the `LinkedBinding` itself as a `Provider` which it is not (providers should link lazily).

### DIFF
--- a/reflect/src/main/java/dagger/reflect/ComponentInvocationHandler.java
+++ b/reflect/src/main/java/dagger/reflect/ComponentInvocationHandler.java
@@ -17,12 +17,12 @@ package dagger.reflect;
 
 import dagger.MembersInjector;
 import dagger.Subcomponent;
+import dagger.reflect.Binding.LinkedBinding;
 import java.lang.reflect.InvocationHandler;
 import java.lang.reflect.Method;
 import java.lang.reflect.Proxy;
 import java.lang.reflect.Type;
 import java.util.concurrent.ConcurrentHashMap;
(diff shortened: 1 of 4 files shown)


reflect/src/main/java/dagger/reflect/ComponentInvocationHandler.java
```diff
@@ -157,12 +157,12 @@ private ComponentInvocationHandler(Scope scope) {
     }
   }
 
-  private static final class SubcomponentBuilderMethodInvocationhandler
+  private static final class SubcomponentBuilderMethodInvocationHandler
```
JakeWharton: Sneaking this in because I'm here

reflect/src/main/java/dagger/reflect/ComponentInvocationHandler.java
```diff
@@ -111,15 +111,15 @@ private ComponentInvocationHandler(Scope scope) {
     @Nullable Object invoke(Object[] args);
   }
 
-  private static final class ProviderMethodInvocationHandler implements MethodInvocationHandler {
-    private final Provider<?> provider;
+  private static final class ProvisionMethodInvocationHandler implements MethodInvocationHandler {
```
JakeWharton: This too...

PR #75, opened 2019-03-17
Add component factory support

JakeWharton:
Subcomponent factories are about 90% supported with this change, but to keep the diff reviewable I'll leave them out for a follow-up. There's also test coverage missing for all the cases where exceptions could be thrown but that can also come later.


JakeWharton:
I was going to rebase this on a refactor, but I'll just rebase my refactor on this.

PR #77, opened 2019-03-18
Extract reflective component scope building to a class

JakeWharton:
This allows the builder, factory, and direct creation to share a bunch of logic. Added `@Module(includes=)` as well which now works for all component creation mechanisms.


integration-tests/src/test/java/com/example/AbstractComponent.java
```diff
@@ -3,7 +3,7 @@
 import dagger.Component;
 
 @Component
-abstract class AbstractComponent {
+public abstract class AbstractComponent {
```
JakeWharton: Visibility now validated earlier and needed to trigger actual desired validation failure

integration-tests/src/test/java/com/example/IntegrationTest.java
```diff
@@ -769,8 +774,9 @@
   @Test public void abstractClassBuilderFails() {
     ignoreCodegenBackend();
 
+    AbstractComponent.Builder builder = backend.builder(AbstractComponent.Builder.class);
     try {
-      backend.builder(AbstractComponent.Builder.class);
+      builder.build();
```
JakeWharton: Now validated later in this call

integration-tests/src/test/java/com/example/IntegrationTest.java
```diff
@@ -848,11 +854,8 @@
     try {
       backend.create(ComponentWithDependencies.class);
       fail();
-    } catch (IllegalArgumentException e) {
-      assertThat(e).hasMessageThat()
-          .isEqualTo("com.example.ComponentWithDependencies "
-              + "declares dependencies [java.lang.String, java.lang.Runnable] "
-              + "and therefore must be created with a builder");
```
JakeWharton: This was a weird special-cased exception. Instead of duplicating for factory, just deleted and relying on other validation.

PR #85, opened 2019-03-26
Add support for @Module(subcomponents=)

JakeWharton:


JakeWharton:
Ugh, rebased hoping it would fix a Dagger functional test but nope. Something else broken 😶 .

PR #100, opened 2019-04-16
Fold BindingMap into Scope and re-work linking

### LINKED ISSUE #97: JIT binding depending on JIT binding fails
I have two classes:
```kotlin
class MyClass1 @Inject constructor(private val delegate: MyClass2)

class MyClass2 @Inject constructor()
```
And the component:
```kotlin
@Singleton
@Component
interface AppComponent {

    val myClass1: MyClass1

    val myClass2: MyClass2
}
```
When I'm trying to provide `MyClass1` like this:
```kotlin
val appComponent = DaggerAppComponent.create()
appComponent.myClass1.doSmth()
```
I got an exception: 
```
java.lang.IllegalStateException: Missing binding for sample.MyClass2
     * Requested: sample.MyClass1
         from @Inject[sample.MyClass1.<init>(…)]
     * Requested: sample.MyClass2
         which was not found.
        at dagger.reflect.Linker.failure(Linker.java:65)
        at dagger.reflect.Linker.get(Linker.java:26)
        at dagger.reflect.UnlinkedJustInTimeBinding.link(UnlinkedJustInTimeBinding.java:46)
        at dagger.reflect.Linker.performLinking(Linker.java:42)
        at dagger.reflect.Linker.link(Linker.java:11)
        at dagger.reflect.Scope.putJitBinding(Scope.java:85)
        at dagger.reflect.Scope.getBinding(Scope.java:33)
        at dagger.reflect.ComponentInvocationHandler.createMethodInvocationHandler(ComponentInvocationHandler.java:111)
        at dagger.reflect.ComponentInvocationHandler.invoke(ComponentInvocationHandler.java:61)
        at java.lang.reflect.Proxy.invoke(Proxy.java:1006)
        at $Proxy4.getMyClass1(Unknown Source)
        ...
```

But when I provide `MyClass2` before providing `MyClass1`:
```kotlin
val appComponent = DaggerAppComponent.create()
appComponent.myClass2
appComponent.myClass1.doSmth()
```
or when I provide it in module:
```kotlin

@Module
object AppModule {

    @Provides
    @JvmStatic
    fun provideMyClass2() = MyClass2()
}
```
then everything is fine

JakeWharton:
This centralizes all of the binding interaction into Scope. As a nice side-effect, JIT bindings depending on JIT bindings can now work with a tiny tweak (that gets admittedly lost in the noise of the diff).

Closes #96. Closes #97.

cc @runningcode


integration-tests/upstream/build.gradle
```diff
@@ -72,8 +67,11 @@ test.filter {
   excludeTest 'dagger.functional.subcomponent.SubcomponentFactoryMethodTest', null
 
   // TODO reflect bug! Members injection on JIT binding does not work.
+  excludeTest 'dagger.functional.cycle.CycleTest', 'lazyIndirectionCycle'
   excludeTest 'dagger.functional.cycle.CycleTest', 'lazyIndirectionSelfCycle'
+  excludeTest 'dagger.functional.cycle.CycleTest', 'providerIndirectionCycle'
   excludeTest 'dagger.functional.cycle.CycleTest', 'providerIndirectionSelfCycle'
+  excludeTest 'dagger.functional.cycle.CycleTest', 'subcomponentIndirectionCycle'
```
JakeWharton: These should all be fixed by #93

reflect/src/main/java/dagger/reflect/Scope.java
```diff
@@ -155,7 +205,14 @@ Builder addBindingIntoSet(Key key, Binding elementBinding) {
         throw new IllegalArgumentException("key.type() must be Set");
       }
 
-      bindings.addIntoSet(key, elementBinding);
+      SetBindings bindings = setBindings.get(key);
```
runningcode: this `bindings` variable shadows the `Builder.bindings`
```suggestion
      SetBindings setBindings = setBindings.get(key);
```

reflect/src/main/java/dagger/reflect/Scope.java
```diff
@@ -198,7 +260,17 @@ Builder addBindingIntoMap(Key key, Object entryKey, Binding entryValueBinding) {
...
-      bindings.addIntoMap(key, entryKey, entryValueBinding);
+      Map<Object, Binding> mapBinding = mapBindings.get(key);
+      //noinspection Java8MapApi Supporting old Android API levels.
+      if (mapBinding == null) {
+        mapBinding = new LinkedHashMap<>();
+        mapBindings.put(key, mapBinding);
+      }
+      Binding replaced = mapBinding.put(entryKey, entryValueBinding);
+      if (replaced != null) {
+        throw new IllegalStateException(); // TODO duplicate keys
```
runningcode: should this throw the same error as on line 183?

integration-tests/src/test/java/com/example/IntegrationTest.java
```diff
@@ -129,8 +129,6 @@
   }
 
   @Test public void justInTimeUnscopedIntoJustInTimeScoped() {
-    ignoreReflectionBackend();
```
runningcode: 👍

reflect/src/main/java/dagger/reflect/Scope.java
```diff
@@ -65,30 +86,52 @@ private Scope(BindingMap bindings, JustInTimeLookup.Factory jitLookupFactory,
...
             : null;
       }
 
       jitBinding = jitBinding.asScoped();
     }
 
-    Binding binding = bindings.putIfAbsent(key, jitBinding);
+    Binding replaced = bindings.putIfAbsent(key, jitBinding);
+    Binding binding = replaced != null
+        ? replaced // You raced another thread and lost.
```
runningcode: just curious, why don't we care about thread races here? because they aren't singletons?

reflect/src/main/java/dagger/reflect/Scope.java
JakeWharton: Yeah it's close. The error message would have to detail that the problem is the `@MapKey` annotation being the problem. Exception messages have mostly been ignored since they can be cleaned up later and don't affect functionality.

reflect/src/main/java/dagger/reflect/Scope.java
JakeWharton: These races are only to create the binding, not the instance.

Two threads might both perform the reflection to find the `@Inject` constructor on some type at the same time. This is the race we don't care about. Once they go to put that binding into our map, that's when we de-duplicate and ensure they agree on who won. This way when you do eventually request an instance we're sure that it only comes from a single, shared binding (which can honor things like `@Singleton`).

JakeWharton: Similarly, we do the same when linking. Two threads might try to link an unlinked binding at the same time. When they go to replace the unlinked binding with the linked one in our map we make sure they agree on who won.

The gist is that it's rare two threads will be doing this work at the same time so we don't try to de-duplicate it. It would be more expensive to have some fancy locking system than just letting two threads do the same work. We just make sure they agree on who won at the end.

reflect/src/main/java/dagger/reflect/Scope.java
JakeWharton: Your suggestion also shadows. It's weird to name the value from a map the same name as the map. The map names might need changed. Technically this one is like `setBindingses`.

PR #101, opened 2019-04-16
Add tests for Lazy and Provider.

runningcode:
References #98

### DIFF
--- a/integration-tests/src/main/java/com/example/LazyScopedBinding.java
+++ b/integration-tests/src/main/java/com/example/LazyScopedBinding.java
@@ -1,0 +1,24 @@
+package com.example;
+
+import dagger.Component;
+import dagger.Lazy;
+import dagger.Module;
+import dagger.Provides;
+
+import javax.inject.Singleton;
+import java.util.concurrent.atomic.AtomicInteger;
+
+@Singleton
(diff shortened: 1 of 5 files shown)


JakeWharton:
I think we want these tests:
 * Provider calls unscoped binding every time `get()` is called, instances are different
 * Provider calls scoped binding once even though `get()` is called twice, instances are same
 * Lazy calls unscoped binding once even though `get()` is called twice, instances are same
 * Two Lazy instances _each_ call unscoped binding once, instances are different
 * Lazy calls scoped binding once when `get()` invoked twice

I would just expose these on the component instead of injecting them into a type like
```java
interface Component {
  Provider<String> string();
}
```

JakeWharton:
There are some checks for provider in the multibindings which do this already which we should copy to keep their structure the same.

TWiStErRob:
Should it be tested that `Lazy` is actually lazy? i.e. the `@Provides` method is not called until `Lazy.get`?

JakeWharton:
Yes. The maps ones have that structure.

runningcode:
Added tests as specified!

JakeWharton:
Thanks! I have this mostly implemented on a branch. Time to rebase and test!

PR #105, opened 2019-05-03
Add support for Provider and Lazy injection

### LINKED ISSUE #98: Add support for javax.inject.Provider and dagger.Lazy

JakeWharton:
Closes #98.


reflect/src/main/java/dagger/reflect/Reflection.java
```diff
@@ -135,9 +135,7 @@ static boolean hasAnnotation(Annotation[] annotations, Class<? extends Annotatio
   }
 
   static void trySet(@Nullable Object instance, Field field, @Nullable Object value) {
-    if ((field.getModifiers() & Modifier.PUBLIC) == 0) {
```
JakeWharton: A public field in a non-public class still needs to be set accessible

PR #106, opened 2019-05-25
Have a slightly better error message when unable to find a jit binding.

runningcode:
This isn't the nice scope chain that we want, but at least it prints our
the missing key, which is bettter than nothing. It also prints out the
linker to know if was null or not.

### DIFF
--- a/reflect/src/main/java/dagger/reflect/Scope.java
+++ b/reflect/src/main/java/dagger/reflect/Scope.java
@@ -69,7 +69,7 @@ final class Scope {
     if (jitLookup != null) {
       LinkedBinding<?> jitBinding = putJitBinding(key, linker, jitLookup);
       if (jitBinding == null) {
-        throw new IllegalStateException(); // TODO nice error message with scope chain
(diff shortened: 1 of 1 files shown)


reflect/src/main/java/dagger/reflect/Scope.java
```diff
@@ -69,7 +69,7 @@ private Scope(ConcurrentHashMap<Key, Binding> bindings, JustInTimeLookup.Factory
     if (jitLookup != null) {
       LinkedBinding<?> jitBinding = putJitBinding(key, linker, jitLookup);
       if (jitBinding == null) {
-        throw new IllegalStateException(); // TODO nice error message with scope chain
+        throw new IllegalStateException("Unable to find binding for key=" + key + " and linker=" + linker); // TODO nice error message with scope chain
```
JakeWharton: Linker has a toString?

runningcode: it doesn't. this error message is mostly to understand if the linker was null or not and what key we were looking for.

PR #107, opened 2019-05-29
Update Dagger, Truth, Gradle, and reflection

JakeWharton:


JakeWharton:
Let me update to 23.0.1 or whatever it is

PR #111, opened 2019-06-10
Add support for multibinding map provider.

runningcode:
Also enable the test for multibinding provider map.

Looking for feedback if this is the right approach to take!


reflect/src/main/java/dagger/reflect/Scope.java
```diff
@@ -58,6 +59,16 @@ private Scope(ConcurrentHashMap<Key, Binding> bindings, JustInTimeLookup.Factory
         Key realKey = Key.of(key.qualifier(), parameterizedKeyType.getActualTypeArguments()[0]);
         return new LinkedLazyBinding<>(this, realKey);
       }
+      if (rawKeyType == Map.class) {
+        if (parameterizedKeyType.getActualTypeArguments()[1] instanceof ParameterizedType) {
```
TWiStErRob: Is there a test for a simple provided `Map`?
```java
@Provides static Map<String, Number> providesMap() {
    Map<String, Number> result = new HashMap<>();
    result.put("one", 1);
    result.put("two", 2.0);
    return result;
}
```
just in case the above condition breaks that.

runningcode: excellent idea! i have made a separate PR for this: https://github.com/JakeWharton/dagger-reflect/pull/112

JakeWharton:
I've been thinking about this for some time, and I may want to think about it longer. But I also don't want to block dagger.android support so if I can't reconcile the idea in my head in a day or two i'll just merge....

reflect/src/main/java/dagger/reflect/LinkedMapBinding.java
```diff
@@ -17,4 +18,28 @@
...
+
+  Binding.LinkedBinding<Map<K, Provider<V>>> asLinkedMapProvidesBinding() {
+    Map<K, LinkedBinding<Provider<V>>> map = new LinkedHashMap<>(entryBindings.size());
+    for (Map.Entry<K, LinkedBinding<V>> elementBinding : entryBindings.entrySet()) {
+      map.put(elementBinding.getKey(), new LinkedProviderBindingWrapper<>(elementBinding.getValue()));
+    }
+    return new LinkedMapProviderBinding<>(map);
+  }
+
+  private final static class LinkedMapProviderBinding<K, V> extends Binding.LinkedBinding<Map<K, Provider<V>>> {
```
runningcode: did a little refactoring to make this an implementation detail instead of a separate class. does that make sense?

JakeWharton:
After some thinking, tell me what you think of this approach:

Change the way maps are bound so that instead of a binding for `Map<K, V>` we put a binding for `Map<K, Provider<V>>` in the graph. Each map then also gets a binding for `Map<K, Lazy<V>` and `Map<K, V>` which each just inject the `Map<K, Provider<V>>` and transform the values into a new map.

What's nice about this approach is that the creation of the `Map<K, Provider<V>>` can be eager inside the linked binding. And then requests for `Map<K, V>` and `Map<K, Lazy<V>>` have to allocate a new map to transform their values, but they already have to do that anyway. And by having bindings for those two transformations eagerly added to the graph (as linked bindings), there's no extraneous allocations.

Does that make sense?

JakeWharton:
I've re-worked this a bit. Basically moved the additional bindings which transform from map of provider to map of lazy or map of value to be inserted into the graph eagerly.

PR #115, opened 2019-06-14
Add test for reusable just in time bindings.

runningcode:


integration-tests/src/main/java/com/example/ReusableScopedJustInTime.java
```diff
@@ -0,0 +1,20 @@
...
+import dagger.Reusable;
+import dagger.Subcomponent;
+
+import javax.inject.Inject;
+
+@Component
+interface ReusableScopedJustInTime {
+  Bar bar();
+
+  @Reusable
```
JakeWharton: Didn't know you could do this

PR #116, opened 2019-06-14
Fix reusable just in time scoping.

runningcode:
Not sure if this is the best way to fix this, but it works. This basically makes the Reusable annotation ignored for the purposes of dagger reflect.

### DIFF
--- a/reflect/src/main/java/dagger/reflect/Reflection.java
+++ b/reflect/src/main/java/dagger/reflect/Reflection.java
@@ -30,6 +30,8 @@ import java.util.LinkedHashSet;
 import java.util.Set;
 import javax.inject.Qualifier;
 import javax.inject.Scope;
+
+import dagger.Reusable;
 import org.jetbrains.annotations.Nullable;
 
 final class Reflection {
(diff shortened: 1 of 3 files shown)


reflect/src/main/java/dagger/reflect/Reflection.java
```diff
@@ -66,9 +68,19 @@
...
+   * Finds scoping annotations that aren't {@link Reusable}.
+   *
+   * @param annotations The set of annotations to parse for scoping annotations.
+   * @return All annotations with Scope or an empty set if none are found.
+   */
   static Set<Annotation> findScopes(Annotation[] annotations) {
     Set<Annotation> scopes = null;
     for (Annotation annotation : annotations) {
+      // Reusable is ignored.
+      if (annotation.annotationType() == Reusable.class) {
```
JakeWharton: Where is the other check for this annotation? I think by moving it here we can delete the other one that already exists.

reflect/src/main/java/dagger/reflect/ReflectiveModuleParser.java
```diff
@@ -82,10 +82,7 @@ private static void addBinding(Scope.Builder scopeBuilder, Key key, Binding bind
       Annotation[] annotations) {
     Annotation scope = findScope(annotations);
     if (scope != null) {
-      if (scope.annotationType().equals(Reusable.class)) {
```
runningcode: @JakeWharton removed it over here!

PR #119, opened 2019-06-22
Add support for generic provider in to a just in time binding.

runningcode:
This allows you to declare and inject following:

```
  final class Thing<T> {
    final Provider<T> genericProvider;

    @Inject Thing(Provider<T> genericProvider) {
      this.genericProvider = genericProvider;
    }
  }
```

This also adds an extra test for multiple parameterized types.

### DIFF
--- a/integration-tests/upstream/build.gradle
+++ b/integration-tests/upstream/build.gradle
@@ -43,14 +43,12 @@ test.filter {
   excludeTest 'dagger.functional.BasicTest', null
 
   // TODO reflect bug! Generics don't work well.
-  excludeTest 'dagger.functional.GenericTest', 'boundedGenerics'
   excludeTest 'dagger.functional.GenericTest', 'complexGenerics'
   excludeTest 'dagger.functional.GenericTest', 'genericModules'
   excludeTest 'dagger.functional.GenericTest', 'membersInjections'
   excludeTest 'dagger.functional.GenericTest', 'noDepsGenerics'
   excludeTest 'dagger.functional.GenericTest', 'packagePrivateTypeParameterDependencies'
   excludeTest 'dagger.functional.GenericTest', 'publicSubclassWithPackagePrivateTypeParameterOfSuperclass'
-  excludeTest 'dagger.functional.GenericTest', 'singletonScopesAppliesToEachResolvedType'
   excludeTest 'dagger.functional.GenericTest', 'testGenericComponentCreate'
   excludeTest 'dagger.functional.GenericTest', 'testGenericDoubleReferences'
   excludeTest 'dagger.functional.GenericTest', 'testGenericSimpleReferences'
--- a/integration-tests/upstream/build.gradle
+++ b/integration-tests/upstream/build.gradle
@@ -43,7 +43,17 @@ test.filter {
   excludeTest 'dagger.functional.BasicTest', null
 
   // TODO reflect bug! Generics don't work well.
-  excludeTest 'dagger.functional.GenericTest', null
+  excludeTest 'dagger.functional.GenericTest', 'boundedGenerics'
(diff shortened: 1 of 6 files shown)


JakeWharton:
Force pushed onto your branch because you fixed one functional test from Dagger proper with this.

I suspect we'll want to factor this logic out so that it can be reused by module parsing which is where the majority of the remaining functional tests are failing.

reflect/src/main/java/dagger/reflect/UnlinkedJustInTimeBinding.java
```diff
@@ -32,7 +43,65 @@
...
+      return parameterizedType;
+    }
+    if (concreteTypeArguments == null) {
+      throw new IllegalStateException(
+          "No concrete type arguments for " + cls + " but needed for " + parameterizedType);
+    }
+    // Iterate through parameterized types declared in this class to find the matching index and
+    // return the corresponding concrete type.
+    for (int i = 0; i < cls.getTypeParameters().length; i++) {
+      if (cls.getTypeParameters()[i] == typeToLookup) {
```
JakeWharton: Is reference equality the right thing to do here? I'm not sure this is guaranteed.

reflect/src/main/java/dagger/reflect/UnlinkedJustInTimeBinding.java
```diff
@@ -32,7 +43,65 @@
...
+   * @return The matching concrete type for the placeholder.
+   */
+  private Type matchingParameterizedType(ParameterizedType parameterizedType) {
+    Type[] actualTypeArguments = parameterizedType.getActualTypeArguments();
+    if (actualTypeArguments.length > 1) {
+      throw new IllegalStateException(
+          "Multiple parameterized type arguments are not supported yet.");
+    }
+    Type typeToLookup = actualTypeArguments[0];
+    // If this isn't a TypeVariableImpl like <T> than we can just return the parameterized type.
```
JakeWharton:
```suggestion
    // If this isn't a TypeVariable like <T> than we can just return the parameterized type.
```

runningcode:
Thanks, I updated the PR with the equals change!

integration-tests/upstream/build.gradle
```diff
@@ -43,7 +43,17 @@ test.filter {
...
+  excludeTest 'dagger.functional.GenericTest', 'complexGenerics'
+  excludeTest 'dagger.functional.GenericTest', 'genericModules'
+  excludeTest 'dagger.functional.GenericTest', 'membersInjections'
+  excludeTest 'dagger.functional.GenericTest', 'noDepsGenerics'
+  excludeTest 'dagger.functional.GenericTest', 'packagePrivateTypeParameterDependencies'
+  excludeTest 'dagger.functional.GenericTest', 'publicSubclassWithPackagePrivateTypeParameterOfSuperclass'
+  excludeTest 'dagger.functional.GenericTest', 'singletonScopesAppliesToEachResolvedType'
+  excludeTest 'dagger.functional.GenericTest', 'testGenericComponentCreate'
+  excludeTest 'dagger.functional.GenericTest', 'testGenericDoubleReferences'
+  excludeTest 'dagger.functional.GenericTest', 'testGenericSimpleReferences'
```
JakeWharton: I'd like to try and see how complicated it is to reduce this list before merging. I fear the current design really only addresses one specific case.

runningcode: I was able to make the solution more generic and have two more tests pass. The rest of the test deal specifically with generic parameters being a part of a Module's inherited declaration! :( (See `ParentModule`)

JakeWharton:
Squashed and merged as e2940ed20bafa2813ed6665f73d5cc57cb3ea806

PR #122, opened 2019-06-25
Make LinkedBinding a Provider

JakeWharton:
This avoids a needless wrapper for map bindings.


JakeWharton:
There's probably other places this can be leveraged but I didn't think to hard about it.

PR #130, opened 2019-07-02
Use annotations instead of methods to ignore integration tests

JakeWharton:
This lets us actually validate that expected-failing reflection tests are actually failing.

### DIFF
--- a/integration-tests/src/test/java/com/example/IntegrationTestRule.java
+++ b/integration-tests/src/test/java/com/example/IntegrationTestRule.java
@@ -1,0 +1,48 @@
+package com.example;
+
+import org.junit.AssumptionViolatedException;
+import org.junit.rules.TestRule;
+import org.junit.runner.Description;
+import org.junit.runners.model.Statement;
+
+final class IntegrationTestRule implements TestRule {
+  private final Backend backend;
+
+  IntegrationTestRule(Backend backend) {
+    this.backend = backend;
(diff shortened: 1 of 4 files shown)


TWiStErRob:
Nice :) Add optional String parameter to explain why it needs to be ignored? Like in the integration Gradle file and built-in `@Ignore` annotation.

integration-tests/src/test/java/com/example/IntegrationTestRule.java
```diff
@@ -0,0 +1,44 @@
...
+  }
+
+  @Override
+  public Statement apply(Statement base, Description description) {
+    boolean ignoreCodegen = description.getAnnotation(IgnoreCodegen.class) != null;
+    if (ignoreCodegen && backend == Backend.CODEGEN) {
+      return new Statement() {
+        @Override
+        public void evaluate() {
+          throw new AssumptionViolatedException("Ignored in code gen backend");
```
runningcode: Could/should we do the same thing as the `@ReflectBug` here and run the test anyways and check if it passes?

JakeWharton: It's not as useful since we don't actually run the Dagger compiler on them. I was actually thinking of splitting them out into their own test since they deal exclusively with error conditions that otherwise happen at compile-time.

Issue #132, opened 2019-07-03
Multibinding with ClassKey cannot be found when using reflect backend


sidius:
Multibindings that use class keys fail to resolve when using reflect backend. 
```
java.lang.IllegalArgumentException: No provider available for java.util.Map<java.lang.Class<?>, com.example.I>
	at dagger.reflect.Scope.getBinding(Scope.java:43)
	at dagger.reflect.ComponentInvocationHandler.createMethodInvocationHandler(ComponentInvocationHandler.java:91)
	at dagger.reflect.ComponentInvocationHandler.invoke(ComponentInvocationHandler.java:56)
	at com.example.$Proxy15.values(Unknown Source)
	at com.example.IntegrationTest.multibindingMapClassKey(IntegrationTest.java:816)
```

Here is the key created by `ReflectiveModuleParser` while adding map bindings: `java.util.Map<java.lang.Class, com.example.I>`

This result in the following map of bindings:
```scope = {Scope@1172} 
 bindings = {ConcurrentHashMap@1181}  size = 3
  0 = {ConcurrentHashMap$MapEntry@1191} "java.util.Map<java.lang.Class, dagger.Lazy<com.example.I>>" -> "Mapjava.util.Map<java.lang.Class, javax.inject.Provider<com.example.I>>"
  1 = {ConcurrentHashMap$MapEntry@1192} "java.util.Map<java.lang.Class, javax.inject.Provider<com.example.I>>" -> "Map{class com.example.Impl1=@Provides[com.example.MultibindingMapClassKey$Module1.one(…)], class com.example.Impl2=@Provides[com.example.MultibindingMapClassKey$Module1.two(…)]}"
  2 = {ConcurrentHashMap$MapEntry@1193} "java.util.Map<java.lang.Class, com.example.I>" -> "Mapjava.util.Map<java.lang.Class, javax.inject.Provider<com.example.I>>"
```
And here is the key used to lookup the binding when constructing `MethodInvocationHandler`:
`java.util.Map<java.lang.Class<?>, com.example.I>`.
Naturally, it fails to resolve the binding and throws an exception.

The fix is rather trivial: use generic return type when constructing keys for map bindings.
See my pr with the test that covers this issue and a fix.

PR #133, opened 2019-07-03
Fix generating bindings for multibindings with class keys.

sidius:
Fixes https://github.com/JakeWharton/dagger-reflect/issues/132

Problem: wrong key for multibindings with class keys. The key has a value of `java.util.Map<java.lang.Class, com.example.I>` when it should be `java.util.Map<java.lang.Class<?>, com.example.I>`.

Solution: fix `ReflectiveModuleParser`: use generic return type when constructing `entryKeyType`.

### DIFF
--- a/reflect/src/main/java/dagger/reflect/ReflectiveModuleParser.java
+++ b/reflect/src/main/java/dagger/reflect/ReflectiveModuleParser.java
@@ -1,6 +1,13 @@
 package dagger.reflect;
 
-import static dagger.reflect.Reflection.*;
+import static dagger.reflect.Reflection.boxIfNecessary;
+import static dagger.reflect.Reflection.findAnnotation;
+import static dagger.reflect.Reflection.findMapKey;
+import static dagger.reflect.Reflection.findQualifier;
+import static dagger.reflect.Reflection.findScope;
+import static dagger.reflect.Reflection.findScopes;
+import static dagger.reflect.Reflection.maybeInstantiate;
+import static dagger.reflect.Reflection.requireAnnotation;
 
 import dagger.Binds;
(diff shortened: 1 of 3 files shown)


reflect/src/main/java/dagger/reflect/ReflectiveModuleParser.java
```diff
@@ -1,13 +1,6 @@
...
 
-import static dagger.reflect.Reflection.boxIfNecessary;
-import static dagger.reflect.Reflection.findAnnotation;
-import static dagger.reflect.Reflection.findMapKey;
-import static dagger.reflect.Reflection.findQualifier;
-import static dagger.reflect.Reflection.findScope;
-import static dagger.reflect.Reflection.findScopes;
-import static dagger.reflect.Reflection.maybeInstantiate;
-import static dagger.reflect.Reflection.requireAnnotation;
+import static dagger.reflect.Reflection.*;
```
runningcode: wildcard imports are not allowed in google java format. i'm surprised this passed verification.

sidius: Yeah, weird that `./gradlew googleJavaFormat` was fine with it. Reverted the change.

JakeWharton:
Thanks!

JakeWharton:
Is this blocking you from using the library? If so I can release a 0.1.1. Otherwise I'll sit on it for a bit until a 0.2.0.

sidius:
> Is this blocking you from using the library? If so I can release a 0.1.1. Otherwise I'll sit on it for a bit until a 0.2.0.

I am using the snapshot version so there is no need to release now.

PR #137, opened 2019-07-10
Add Lint Rules for missing or wrong RetentionType

winterDroid:
@runningcode mentioned in his DroidCon Berlin talk that some Lint rules would be beneficial
This PR adds the first two:
- MissingRetention: Qualifier or MapKey annotation did not specify a Retention
- WrongRetention: Qualifier or MapKey annotation did specify a Retention other than RUNTIME


reflect-lint/gradle.properties
```diff
@@ -0,0 +1,3 @@
+POM_NAME=Lint Rules for Dagger Reflect
+POM_ARTIFACT_ID=dagger-reflect-lint
+POM_PACKAGING=jar
```
runningcode: should this be a `jar` or `aar`? or is that not relevant for the pom?

reflect-lint/src/main/java/dagger/reflect/lint/IssueRegistry.java
```diff
@@ -0,0 +1,16 @@
+package dagger.reflect.lint;
+
+import com.android.tools.lint.detector.api.Issue;
+import java.util.Arrays;
+import java.util.List;
+import org.jetbrains.annotations.NotNull;
+
+public class IssueRegistry extends com.android.tools.lint.client.api.IssueRegistry {
```
runningcode: change the name so we can static import?

reflect-lint/gradle.properties
winterDroid: By using a `jar` it should theoretically be possible to use it with non-Android projects and the standalone Lint version.

reflect-lint/src/main/java/dagger/reflect/lint/IssueRegistry.java
winterDroid: Do you have a suggestion for the name? `DaggerReflectIssueRegistry`?

TWiStErRob:
@winterDroid which talk?

reflect-lint/gradle.properties
JakeWharton: It always has to be a jar, even when embedding in aars or depending via the `lintChecks` configuration

build.gradle
```diff
@@ -3,7 +3,8 @@ import net.ltgt.gradle.errorprone.CheckSeverity
 buildscript {
   ext.versions = [
       'dagger': '2.23.2',
-      'incap' : '0.2'
+      'incap': '0.2',
+      'androidTools': '26.4.1'
```
JakeWharton:
```suggestion
      'androidTools': '26.4.1',
```

build.gradle
```diff
@@ -38,6 +39,12 @@ buildscript {
               'core': 'androidx.test:core:1.2.0',
           ],
       ],
+      'lint': [
+          'core': "com.android.tools.lint:lint:${versions.androidTools}",
+          'api': "com.android.tools.lint:lint-api:${versions.androidTools}",
+          'checks': "com.android.tools.lint:lint-checks:${versions.androidTools}",
+          'tests': "com.android.tools.lint:lint-tests:${versions.androidTools}"
```
JakeWharton:
```suggestion
          'tests': "com.android.tools.lint:lint-tests:${versions.androidTools}",
```

reflect-lint/src/main/java/dagger/reflect/lint/DaggerReflectIssueRegistry.java
```diff
@@ -0,0 +1,18 @@
+package dagger.reflect.lint;
+
+import com.android.tools.lint.client.api.IssueRegistry;
+import com.android.tools.lint.detector.api.Issue;
+import org.jetbrains.annotations.NotNull;
+
+import java.util.Arrays;
+import java.util.List;
+
+public class DaggerReflectIssueRegistry extends IssueRegistry {
```
JakeWharton:
```suggestion
public final class DaggerReflectIssueRegistry extends IssueRegistry {
```

reflect-lint/src/main/java/dagger/reflect/lint/WrongRetentionDetector.java
```diff
@@ -0,0 +1,189 @@
...
+      final String kotlinQualifiedName = CLASS_KOTLIN_RETENTION_POLICY + "." + policy.name();
+      if (javaQualifiedName.equals(retentionPolicy)
+          || kotlinQualifiedName.equals(retentionPolicy)) {
+        return policy.name();
+      }
+    }
+    return null;
+  }
+
+  public static final Issue ISSUE_MISSING_RETENTION =
```
JakeWharton: I don't think it's useful to have these as two checks instead of one. When no annotation is specified there is a default retention (of class) so it's really the same situation as having the wrong value. Even if that weren't true, I don't think the distinction is that useful.

winterDroid:
> @winterDroid which talk?

You can find more information on the [speaker profile](https://www.de.droidcon.com/speaker/Nelson-Osacky) and on the [schedule](https://www.de.droidcon.com/schedule) itself. Unfortunately I was not able to link to the talk directly. AFAIK the recordings are not yet published. But [here](https://speakerdeck.com/runningcode/dagger-reflect-the-circle-from-runtime-to-compile-time-and-back-again) you can find the slides.

reflect-lint/src/main/java/dagger/reflect/lint/WrongRetentionDetector.java
winterDroid: Good point. I removed the separate issue but kept the logic the same such that the error messages are still different.

reflect-lint/build.gradle
```diff
@@ -0,0 +1,20 @@
+apply plugin: 'java-library'
+
+targetCompatibility = JavaVersion.VERSION_1_8
+sourceCompatibility = JavaVersion.VERSION_1_8
+
+dependencies {
+  compileOnly deps.lint.api
+  compileOnly deps.lint.checks
+
+  testImplementation deps.lint.core
```
jrodbx: what's `core` needed for?

winterDroid: Oh yeah, actually that's not necessary...

TWiStErRob:
Finally found the full talk at https://www.droidcon.com/videos?path=Berlin/2019:

Dagger Reflect - The circle from runtime to compile time and back to runtime
https://www.droidcon.com/media-detail?video=357540042
(not on YouTube yet)

Issue #136, opened 2019-07-09
Using component as a dependency


Laimiux:
I'm curious if there are plans to add support for the following use case. I'm using the root component as an injected parameter in the same graph. For example: 
```kotlin
@Module
object AppModule {
    @Provides fun provideFeatureViewModel(component: AppComponent) 
        = Feature.create(component) 
}
```

A bit more complete example:
```kotlin
object Feature {
    interface Dependencies {
        fun foo(): Foo
    }

    @Component(dependencies = [Dependencies::class])
    interface FeatureComponent {
        fun featureViewModel(): FeatureViewModel
    }

    fun create(dependencies: Dependencies): FeatureViewModel {
        return TODO()
    }
}

@Component(modules = [AppModule::class])
interface AppComponent : Feature.Dependencies {

}

@Module
object AppModule {
    @Provides fun provideFeatureViewModel(component: AppComponent) 
        = Feature.create(component) 
}
```

JakeWharton:
So you're saying that Dagger exposes a binding for the component instance inside that component automatically? And presumably this works for subcomponents as well?

Laimiux:
Correct, though I'm not sure if parent component binding is available within the subcomponent graph.

JakeWharton:
Should be easy to test for!

Laimiux:
Added a test for this functionality #138. I've looked a little bit into the code base to add this functionality. Given the sort of cyclical nature of this, it's not completely obvious how to best add it. Maybe you already have ideas?

JakeWharton:
My current thinking here is that during scope construction we can put a semi-mutable binding into the graph for the (sub)component type and then sneak the instance into the binding after the scope and (sub)component has been created.

Laimiux:
Sounds reasonable, will take a look at implementing it.

PR #138, opened 2019-07-10
Component binding instance functionality test.

Laimiux:
Currently, this PR only adds a test for this functionality. As expected, the `reflect` version fails.

Reference: #136

### DIFF
--- a/integration-tests/src/test/java/com/example/IntegrationTest.java
+++ b/integration-tests/src/test/java/com/example/IntegrationTest.java
@@ -1345,7 +1345,7 @@ public final class IntegrationTest {
   @Test
   public void componentBindingInstance() {
     ComponentBindingInstance instance = backend.create(ComponentBindingInstance.class);
-    assertThat(instance).isEqualTo(instance.result().foo);
+    assertThat(instance).isSameAs(instance.result().foo);
   }
 
   @Test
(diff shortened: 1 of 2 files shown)


integration-tests/src/test/java/com/example/IntegrationTest.java
```diff
@@ -1342,6 +1342,12 @@ public void subcomponentFactoryProvision() {
     assertThat(nested.two()).isEqualTo(2L);
   }
 
+  @Test
+  public void componentBindingInstance() {
+    ComponentBindingInstance instance = backend.create(ComponentBindingInstance.class);
+    assertThat(instance).isEqualTo(instance.result().foo);
```
JakeWharton:
```suggestion
    assertThat(instance).isSameAs(instance.result().foo);
```

integration-tests/src/test/java/com/example/IntegrationTest.java
```diff
@@ -1342,6 +1342,12 @@ public void subcomponentFactoryProvision() {
     assertThat(nested.two()).isEqualTo(2L);
   }
 
+  @Test
```
JakeWharton: Can you add `@ReflectBug` to this so we can merge it?

JakeWharton:
Merged as 29b829d. Thanks!

PR #140, opened 2019-07-11
Process included modules when parsing `ContributesAndroidInjector` modules

amatkovsky:
`ReflectiveModuleParser` only takes into account the modules directly specified in `ContributesAndroidInjector` annotation and ignores included modules.

Solution: go through the modules recursively and collect all included modules ignoring the duplicates to avoid duplicate bindings.


JakeWharton:
Hmm we have this code more than once now. It feels like the responsibility to do this should be centralized _somewhere_ rather than requiring everyone to do it themselves. I'll take a closer look soon.

JakeWharton:
I've reworked this in #152 to share the logic with the other code which handles modules.

PR #141, opened 2019-07-11
Add component and subcomponent instance binding.

Laimiux:
Changes:
- Added functionality
- Removed `@ReflectBug`
- Added a test for subcomponent instance binding.

Related to: #136


JakeWharton:
I've reworked this a bit in #151 to prevent `Scope` and `Scope.Builder` from knowing anything about components or being mutable.

PR #143, opened 2019-07-14
Add R8/ProGuard rules to the codegen artifact

JakeWharton:
Also process the codegen test code with both R8 and ProGuard to ensure they work correctly in practice.

### DIFF
--- a/codegen/src/main/resources/META-INF/proguard/dagger-reflect.pro
+++ b/codegen/src/main/resources/META-INF/proguard/dagger-reflect.pro
@@ -1,0 +1,49 @@
+# When using factories or builders, the enclosing class is used to determine the component type.
+-keepattributes InnerClasses
+-keepattributes EnclosingMethod # Required by InnerClasses
+
+# JSR 330 and Dagger annotations like @Inject, @Provide, @Component, etc. are obviously needed.
+-keepattributes RuntimeVisibleAnnotations
+
+# Annotation defaults are required for @MapKey, @Component, @Subcomponent, etc.
+-keepattributes AnnotationDefault
+
+# Generic signatures are needed for properly parsing types of module methods, component methods,
+# and injected fields/methods.
+-keepattributes Signature
+
+# The names of component and subcomponent types must be kept so the Dagger-generated type can be
+# resolved by prepending a prefix to the name. This should also prevent vertical class merging.
+-keepnames @dagger.Component class *
+-keepnames @dagger.Subcomponent class *
+
+# For each component type, keep the corresponding Dagger-generated type. We also keep the create
+# method, in case there is no builder or factory.
+-if @dagger.Component class **.*
+-keep class <1>.Dagger<2> {
+  static ** create();
+}
+
+# For each component builder type that is kept, keep the name of that type which will ensure it
+# remains nested inside its corresponding component type.
+-if @dagger.Component$Builder class **.*$*
+-keep class <1>.<2>
+
+# For each component builder type that is kept, keep the corresponding Dagger-generated component
+# type and the factory method for the builder.
+-if @dagger.Component$Builder class **.*$*
+-keep class <1>.Dagger<2> {
+  static ** builder();
+}
+
+# For each component factory type that is kept, keep the name of that type which will ensure it
+# remains nested inside its corresponding component type.
+-if @dagger.Component$Factory class **.*$*
+-keep class <1>.<2>
+
+# For each component factory type, keep the corresponding Dagger-generated component type and the
+# factory method for the factory.
+-if @dagger.Component$Factory class **.*$*
+-keep class <1>.Dagger<2> {
+  static ** factory();
+}
--- a/codegen/src/test/rules.pro
+++ b/codegen/src/test/rules.pro
@@ -1,0 +1,18 @@
+# @org.junit.Test
+-keepattributes RuntimeVisibleAnnotations
+
+# Test methods are invoked reflectively.
+-keepclasseswithmembers class * {
+  @org.junit.Test <methods>;
+}
+
+# Test classes are constructed reflectively.
+-keepclassmembers class **.*Test {
+  <init>(...);
+}
+
+# Annotation to preserve conditions required for tests which otherwise aren't automatically kept.
+-keep @dagger.Keep class *
+
+# For Mac OS and it's case-insensitive (by default) filesystem.
+-dontusemixedcaseclassnames
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,8 @@ language: android
 
 jdk:
   - oraclejdk8
-  - openjdk11
+  # TODO Java 11 is currently disabled because R8 and ProGuard (in codegen build) require rt.jar
+  #- openjdk11
 
 before_install:
   # Install SDK license so Android Gradle plugin can install deps.
--- a/codegen/build.gradle
+++ b/codegen/build.gradle
@@ -1,13 +1,118 @@
+import org.gradle.internal.jvm.Jvm
+
 apply plugin: 'java-library'
 
 sourceCompatibility = JavaVersion.VERSION_1_8
 targetCompatibility = JavaVersion.VERSION_1_8
+
+configurations {
+  r8
+  proGuard
+}
+
+// TODO Remove when R8 is updated to 1.5 or newer which pulls kotlinx.metadata from Maven Central.
+repositories {
+  maven {
+    url 'https://kotlin.bintray.com/kotlinx'
+  }
+}
 
 dependencies {
   testImplementation deps.junit
   testImplementation deps.truth
   testImplementation deps.dagger.runtime
   testAnnotationProcessor deps.dagger.compiler
+
+  r8 'com.android.tools:r8:1.4.93'
+  proGuard 'net.sf.proguard:proguard-base:6.1.1'
 }
+
+def r8Classes = new File(buildDir, 'classes/java/testR8')
+task runR8(type: JavaExec, dependsOn: 'testClasses') {
+  classpath = configurations.r8
+  main = 'com.android.tools.r8.R8'
+
+  doFirst {
+    if (r8Classes.exists()) {
+      r8Classes.deleteDir()
+    }
+    r8Classes.mkdirs()
+
+    def arguments = [
+        '--release',
+        '--classfile',
+        '--output', r8Classes.absolutePath,
+        '--pg-conf', file('src/test/rules.pro').absolutePath,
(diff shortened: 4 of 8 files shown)


codegen/src/main/resources/META-INF/proguard/dagger-reflect.pro
```diff
@@ -0,0 +1,49 @@
+# When using factories or builders, the enclosing class is used to determine the component type.
```
TWiStErRob: I don't see a mention of `@Binds` which I think may be a special case as it's not referenced by generated Dagger component code. Is it covered by one of these rules?

JakeWharton: These rules are for the codegen artifact reflection, not the reflection in the reflect artifact.

I'm not convinced anyone will be using shrinking with the reflect artifact. I'd be curious what their use case for this was.

codegen/src/test/rules.pro
```diff
@@ -0,0 +1,18 @@
...
+
+# Test classes are constructed reflectively.
+-keepclassmembers class **.*Test {
+  <init>(...);
+}
+
+# Annotation to preserve conditions required for tests which otherwise aren't automatically kept.
+-keep @dagger.Keep class *
+
+# For Mac OS and it's case-insensitive (by default) filesystem.
```
JakeWharton:
```suggestion
# For Mac OS and its case-insensitive (by default) filesystem.
```

codegen/src/main/resources/META-INF/proguard/dagger-reflect.pro
TWiStErRob: Oh, good point, sorry.

Re shrinking with reflect: we sometimes minifyEnabled=true on debug flavor so we can test shrinking-related issues. If dagger-reflect was set up it with `if (idea)` this quick-change-for-debugging would fail the build unexpectedly because of reflect.

JakeWharton: In that case, though, having dagger-reflect in your APK would already produce results that were not representative of what happens when shrinking in release. In release builds a large portion of Dagger-generated types are vertically merge and/or inlined. I've even written some non-trivial graphs that have been _completely_ eliminated by shrinking leaving only a bunch of chained calls to constructors.

I'm not opposed to adding rules to the reflect artifact. They're probably pretty basic. But I think the concept of shrinking and using reflect are mutually exclusive in their goals so it's not a high priority.

Laimiux:
The ruleset crashed on my branch using the activity in `integration-tests:android`. https://github.com/Laimiux/dagger-reflect/tree/laimonas/r8_rule_test.

To replicate the issue:
1. I've replaced `proguard-rules.pro` with the new rules in this PR
2. Ran `./gradlew :integration-tests:android:installCodegenDebug`

JakeWharton:
Thanks. I'll look soon. Although I'd be happy to land this as-is and fix that separately as a bug.

Laimiux:
Using this ruleset even basic use of `AppComponent` crashes. I suspect that the R8 doesn't run correctly during the test you have written because that exact use case crashes for me.

Laimiux:
I might be misunderstanding how it works, do we need to mark components with `@Keep`?

JakeWharton:
When I wrote them I was inspecting the classfile output of both tools and
both looked correct.

Laimiux:
Any ideas of what could be the cause for the crash?

JakeWharton:
I haven't had time to look at it yet so I don't even know what's crashing.

TWiStErRob:
@JakeWharton Played around a bit, but couldn't make it work.

1. In the current state applying the proguard rules, DaggerAppCompnent is not kept:
```gradle
  buildTypes {
    debug {
      minifyEnabled true
      shrinkResources minifyEnabled
    }
  }
```
```
     at android.app.ActivityThread.handleBindApplication(ActivityThread.java:5876)
        at android.app.ActivityThread.access$1100(ActivityThread.java:199)
        at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1650)
        at android.os.Handler.dispatchMessage(Handler.java:106)
        at android.os.Looper.loop(Looper.java:193)
        at android.app.ActivityThread.main(ActivityThread.java:6669)
        at java.lang.reflect.Method.invoke(Native Method)
        at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:493)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:858)
     Caused by: java.lang.IllegalStateException: Unable to find generated component implementation com.example.DaggerAppComponent for component com.example.AppComponent
        at a.b.a.c(:61)
        at a.b.a.a(:25)
        at a.a.a(:22)
        at com.example.ExampleApp.onCreate(:14)
        at android.app.Instrumentation.callApplicationOnCreate(Instrumentation.java:1154)
        at android.app.ActivityThread.handleBindApplication(ActivityThread.java:5871)
        at android.app.ActivityThread.access$1100(ActivityThread.java:199) 
        at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1650) 
        at android.os.Handler.dispatchMessage(Handler.java:106) 
        at android.os.Looper.loop(Looper.java:193) 
        at android.app.ActivityThread.main(ActivityThread.java:6669) 
        at java.lang.reflect.Method.invoke(Native Method) 
        at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:493) 
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:858) 
     Caused by: java.lang.ClassNotFoundException: Didn't find class "com.example.DaggerAppComponent" on path: DexPathList[[zip file "/data/app/com.example-cfNVrdQea-4kuqzGxdZGhg==/base.apk"],nativeLibraryDirectories=[/data/app/com.example-cfNVrdQea-4kuqzGxdZGhg==/lib/x86_64, /system/lib64]]
        at dalvik.system.BaseDexClassLoader.findClass(BaseDexClassLoader.java:134)
        at java.lang.ClassLoader.loadClass(ClassLoader.java:379)
        at java.lang.ClassLoader.loadClass(ClassLoader.java:312)
        at a.b.a.c(:54)
        at a.b.a.a(:25) 
        at a.a.a(:22) 
        at com.example.ExampleApp.onCreate(:14) 
        at android.app.Instrumentation.callApplicationOnCreate(Instrumentation.java:1154) 
        at android.app.ActivityThread.handleBindApplication(ActivityThread.java:5871) 
        at android.app.ActivityThread.access$1100(ActivityThread.java:199) 
        at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1650) 
        at android.os.Handler.dispatchMessage(Handler.java:106) 
        at android.os.Looper.loop(Looper.java:193) 
        at android.app.ActivityThread.main(ActivityThread.java:6669) 
        at java.lang.reflect.Method.invoke(Native Method) 
        at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:493) 
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:858) 
```

2. The rule looks correct, but doesn't keep:

```proguard
-whyareyoukeeping class com.example.AppComponent
-whyareyoukeeping class com.example.DaggerAppComponent
```

```properties
android.enableR8=true

com.example.AppComponent
|- is referenced in keep rule:
|  <PROGUARD_COMPATIBILITY_RULE>:Unknown
Nothing is keeping com.example.DaggerAppComponent
```

```properties
android.enableR8=true
android.enableR8.fullMode=true

Nothing is keeping com.example.AppComponent
Nothing is keeping com.example.DaggerAppComponent
```

```properties
android.enableR8=false

com.example.AppComponent
  is invoked by    com.example.DaggerAppComponent.create (32:32)
  is kept by a directive in the configuration.

com.example.DaggerAppComponent
  is kept by a directive in the configuration.

# but still crashes with java.lang.ClassNotFoundException: Didn't find class "com.example.Daggera"
```

3. Whenever I looked in the APK classes.dex, it matched the `whyareyoukeeping` output.

mochadwi:
> 3. Manually adding the rules, everything works with all variants of shrinking:
> 
> ```
> -keepnames class com.example.AppComponent
> -keep class com.example.DaggerAppComponent {
>   static ** create();
> }
> ```
> 
> 4. Whenever I looked in the APK classes.dex, it matched the `whyareyoukeeping` output.

You stated `works` and `matched` @TWiStErRob, did the point 3 & 4 solved the crashing issue?

TWiStErRob:
Uuuuh, @mochadwi I have no memory of doing this investigation :D, but here are my thoughts:
3. works: I said "everything works", from me that usually means app works as expected, no warnings, no errors, no crashes, just works.
4. matched: I validated whether ProGuard/r8's `-whyareyoukeeping` log output matches what's actually bundled in the `classes.dex` inside the minified APK.

The whole point of the comment is validating @Laimiux's claim that in the current state it crashes, and trying to find out why to help @JakeWharton, which is: the keeps to achieve what's in 3. is missing (or handled differently in all 3 shrinking modes) from the generic rules in this PR.

So of course, manually keeping your `@Component`s and their reflectively accessed generated implementations works as expected, but the goal of this PR is to make it work in a general case without having to add our entry points manually. to our ProGuard rules.

codegen/src/main/resources/META-INF/proguard/dagger-reflect.pro
TWiStErRob: What I was referring to here is adding `minifyEnabled=true` to debug so that we can verify things not related to Dagger, or dagger-reflect. But at the moment if the debug build is set up with dagger-reflect, it prevents debugging other problems., because injection crashes. I see why it's not high priority though, as it shouldn't be hard to turn off dagger-reflect, it's just inconvenient.

mochadwi:
> Uuuuh, @mochadwi I have no memory of doing this investigation :D, but here are my thoughts:
> 3. works: I said "everything works", from me that usually means app works as expected, no warnings, no errors, no crashes, just works.
> 4. matched: I validated whether ProGuard/r8's `-whyareyoukeeping` log output matches what's actually bundled in the `classes.dex` inside the minified APK.
> 
> The whole point of the comment is validating @Laimiux's claim that in the current state it crashes, and trying to find out why to help @JakeWharton, which is: the keeps to achieve what's in 3. is missing (or handled differently in all 3 shrinking modes as seen in 2.) from the generic rules in this PR.
> 
> So of course, manually keeping your `@Component`s and their reflectively accessed generated implementations work as expected, but the goal of this PR is to make it work in a general case (only 1.) without having to add our entry points manually (3.). to our ProGuard rules.

Nice, I really appreciate it for your answer. That's explained everything about this issue still open. @TWiStErRob

PR #144, opened 2019-07-16
Add test for nullability mismatch.

runningcode:
When there is a mismatch in nullability between the binding and the request, dagger
throws an exception during compile time. We should throw the same
exception at runtime for Dagger Reflect.

The message should be something like:
```String is not nullable, but is being provided by @Provides @Nullable String com.example.BindsProviderNull.Module1.provideString()```


runningcode:
Curious if you have any ideas on solving this. Should the nullability annotation be a part of `Key`?

JakeWharton:
It's not part of the key because you can't have a nullable and non-nullable binding for the same type. I think it might be a boolean on the binding that the requestor and binding need to agree on.

PR #149, opened 2019-07-19
Add check for nullability mismatch.

runningcode:
When there is a mismatch between the nullability of the binding and the
requestor, throw an exception.

This only works if the nullability annotations have runtime retention.

Looking for feedback to see if this is a reasonable approach and how/if we should make the error message better.


reflect/src/main/java/dagger/reflect/Reflection.java
```diff
@@ -121,6 +121,19 @@ static boolean hasAnnotation(Annotation[] annotations, Class<? extends Annotatio
     return findAnnotation(annotations, annotation) != null;
   }
 
+  /**
+   * Check for any annotation which is named "Nullable". Warning: many Nullable annotations do not
+   * have runtime retention.
+   */
+  static boolean hasNullableAnnotation(Method method) {
+    for (Annotation annotation : method.getDeclaredAnnotations()) {
+      if (annotation.annotationType().getName().contains("Nullable")) {
```
TWiStErRob: https://github.com/google/dagger/blob/c436b390d87b2e64b3e39c2fcdcb5a1f84dd34eb/java/dagger/internal/codegen/ConfigurationAnnotations.java#L82

reflect/src/main/java/dagger/reflect/ComponentInvocationHandler.java
```diff
@@ -123,13 +124,22 @@ private ComponentInvocationHandler(Scope scope) {
...
 
-    ProvisionMethodInvocationHandler(LinkedBinding<?> binding) {
+    ProvisionMethodInvocationHandler(LinkedBinding<?> binding, boolean nullable) {
       this.binding = binding;
+      this.nullable = nullable;
     }
 
     @Override
     public @Nullable Object invoke(Object[] args) {
+      if (binding instanceof LinkedProvidesBinding) {
```
TWiStErRob: `@BindsInstance` could also have `@Nullable` can't it?

runningcode:
Hmm, should we introduce `DependencyRequest` to dagger reflect?

reflect/src/main/java/dagger/reflect/ComponentInvocationHandler.java
JakeWharton: Yeah I think there needs to be something like a `isNullable` on `Binding` until type annotations save us from this nonsense. And then anyone requesting a `Binding` will have to validate that the consumer can handle null.

PR #150, opened 2019-07-22
[bugfix] getDistinctTypeHierarchy didn't expand interface hierarchy.

Laimiux:
Reflection.getDistinctTypeHierarchy now handles the following use case.
```kotlin
interface First extends Second {}
interface Second extends Third {}
interface Third {}
```

### DIFF
--- a/integration-tests/src/test/java/com/example/NestedDependencyInterfaceTest.java
+++ b/integration-tests/src/test/java/com/example/NestedDependencyInterfaceTest.java
@@ -1,0 +1,25 @@
+package com.example;
+
+import dagger.Component;
+
+@Component(dependencies = NestedDependencyInterfaceTest.First.class)
+public interface NestedDependencyInterfaceTest {
+  String value();
+
+  @Component.Factory
+  interface Factory {
+    NestedDependencyInterfaceTest create(First dependencies);
+  }
+
+  interface First extends Second {
+
+  }
+
+  interface Second extends Third {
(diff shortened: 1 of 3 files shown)


JakeWharton:
Can we test for this? Otherwise it'll likely regress. Maybe a module with that interface hierarchy and a static provider on `Third`.

Laimiux:
Added a test, however `code gen` variant fails because it cannot find `DaggerNestedDependencyInterfaceTest` class.

integration-tests/src/test/java/com/example/NestedDependencyInterfaceTest.java
```diff
@@ -0,0 +1,25 @@
+package com.example;
+
+import dagger.Component;
+
+@Component(dependencies = NestedDependencyInterfaceTest.First.class)
+public interface NestedDependencyInterfaceTest {
```
TWiStErRob: `DaggerNestedDependencyInterfaceTest` is generated by traditional dagger, which is only executed on production code:
https://github.com/JakeWharton/dagger-reflect/blob/b009ca8d3cdd6dac0a6c653ea4459e02996a540e/integration-tests/build.gradle#L8

Move this class to `src/main/java`.

JakeWharton:
Rebased, moved the test subject to `src/main/`, removed `@IgnoreCodegen`, and merged. Thanks!

PR #154, opened 2019-07-30
Improve error message for incorrect return type with BindsIntoSet.

runningcode:
There were also duplicate error messages so I removed the redundant
error message and made it more clear exactly what the error is.

### DIFF
--- a/reflect/src/main/java/dagger/reflect/Scope.java
+++ b/reflect/src/main/java/dagger/reflect/Scope.java
@@ -241,7 +241,7 @@ final class Scope {
     }
 
     /**
-     * Adds a new element into the set specified by {@code key}.
+     * Adds a new set of elements into the set specified by {@code key}.
      *
      * @param key The key defining the set into which this element will be added. The raw class of
      *     the {@linkplain Key#type() type} must be {@link Set Set.class}.
--- a/integration-tests/src/test/java/com/example/BindsElementsIntoSetWrongReturn.java
+++ b/integration-tests/src/test/java/com/example/BindsElementsIntoSetWrongReturn.java
@@ -1,0 +1,29 @@
+package com.example;
+
+import dagger.Binds;
+import dagger.Component;
+import dagger.Module;
+import dagger.Provides;
+import dagger.multibindings.ElementsIntoSet;
+import java.util.ArrayDeque;
+import java.util.Deque;
+import java.util.Set;
+
+@Component(modules = BindsElementsIntoSetWrongReturn.Module1.class)
+interface BindsElementsIntoSetWrongReturn {
+  Set<String> strings();
+
+  @Module
+  abstract class Module1 {
+    @Provides
+    static Deque<String> strings() {
+      ArrayDeque<String> strings = new ArrayDeque<>();
+      strings.add("foo");
+      return strings;
+    }
+
+    @Binds
+    @ElementsIntoSet
+    abstract String setStrings(Deque<String> foo);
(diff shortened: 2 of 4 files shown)


reflect/src/main/java/dagger/reflect/Scope.java
```diff
@@ -252,7 +252,7 @@ Builder addBindingElementsIntoSet(Key key, Binding elementsBinding) {
       if (key == null) throw new NullPointerException("key == null");
       if (elementsBinding == null) throw new NullPointerException("elementsBinding == null");
       if (Types.getRawType(key.type()) != Set.class) {
-        throw new IllegalArgumentException("key.type() must be Set");
+        throw new IllegalArgumentException("@BindsIntoSet must return Set");
```
JakeWharton: I have tentative plans to expose `Scope` as public API. It shouldn't know anything about Dagger and should function like a general-purpose JSR 330-ish service locator. So I would prefer to keep this exception message the same (or, well, it should really print the offending key...).

runningcode: Ah, I'll update it to print the key. So duplicate the exception in both places?
I wanted to add another check here in a follow up PR similar to the check in Dagger which says `@Binds methods must return a primitive, an array, a type variable, or a declared type` if the return type is `Set<? extends String?`.

Should we move the check out to a common place? Or will your upcoming changes to expose Scope as a public API move these checks somewhere else?

JakeWharton: Yeah duplicate it. Think about it like this: Scope is validating what it knows about which is only keys mapped to bindings and it wants to prevent a duplicate. `@BindsIntoSet` is validating a method shape for that specific annotation in order to create a key and binding to insert into a scope.

I'll try to pull out the non-Dagger-specific bits tomorrow or next week. Any naming suggestions? It's completely Dagger-free so it can be anything.

runningcode: What about Sheath? It's where you stick your daggers!

JakeWharton: That used to be a thing for getting Dagger running on GWT: https://github.com/tbroyer/sheath. Seems like it could be used.

reflect/src/main/java/dagger/reflect/Scope.java
```diff
@@ -252,7 +252,8 @@ Builder addBindingElementsIntoSet(Key key, Binding elementsBinding) {
       if (key == null) throw new NullPointerException("key == null");
       if (elementsBinding == null) throw new NullPointerException("elementsBinding == null");
       if (Types.getRawType(key.type()) != Set.class) {
-        throw new IllegalArgumentException("key.type() must be Set");
+        throw new IllegalArgumentException(
+            "@BindsIntoSet must return Set. Found " + key.type() + ".");
```
JakeWharton: This still need reverted

runningcode: you mean keep it as `throw new IllegalArgumentException("key.type() must be Set");`?
isn't this always used in the case of `@BindsIntoSet`

JakeWharton: Yes. And yes, but `Scope` shouldn't know anything about how Dagger works.

runningcode: Done. Also updated the javadoc on this method.

PR #156, opened 2019-08-03
Add Linker and Scope toString + tests

TWiStErRob:
Fixes:
```
reflect\src\main\java\dagger\reflect\Scope.java:84:
warning: [ObjectToString] Linker is final and does not override Object.toString, so converting it to a string will print its identity (e.g. `Linker@ 4488aabb`) instead of useful information
.
            "Unable to find binding for key=" + key + " with linker=" + linker);
                                                                        ^
    (see https://errorprone.info/bugpattern/ObjectToString)
```

### DIFF
--- a/integration-tests/src/test/java/com/example/IntegrationTest.java
+++ b/integration-tests/src/test/java/com/example/IntegrationTest.java
@@ -245,7 +245,11 @@ public final class IntegrationTest {
       component.thing();
       fail();
     } catch (IllegalStateException e) {
-      // TODO assert some message
+      assertThat(e)
+          .hasMessageThat()
+          .isEqualTo(
+              "Unable to find binding for key=com.example.JustInTimeWrongScope$Thing"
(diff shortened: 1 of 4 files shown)


integration-tests/src/test/java/com/example/IntegrationTest.java
```diff
@@ -158,6 +158,24 @@ public void optionalBindingNullable() {
     }
   }
 
+  @Test
+  @IgnoreCodegen
```
TWiStErRob: This is actually compiling and working in codegen, but this test would fail at `fail()` with codegen.
I couldn't find any repro other than `OptionalBindingWrongScope` where linker was not null, any ideas?

JakeWharton: I merged the first two commits while I think about this more.

I didn't realize qualifier annotations on types actually worked...

PR #160, opened 2019-08-05
Improve error message for incorrect return types.

runningcode:
This improves the error message for more cases when using `@BindsIntoSet`.

This copies the error message from dagger.

See https://github.com/google/dagger/blob/f9e41bc94701c52c60c534390b59b135444276a4/java/dagger/internal/codegen/BindingElementValidator.java#L101

### DIFF
--- a/integration-tests/src/test/java/com/example/IntegrationTest.java
+++ b/integration-tests/src/test/java/com/example/IntegrationTest.java
@@ -108,6 +108,21 @@ public final class IntegrationTest {
   }
 
   @Test
+  @IgnoreCodegen
+  public void bindElementsIntoSetGenericWrongReturn() {
+    try {
+      backend.create(BindsElementsIntoSetGenericWrongReturn.class);
+      fail();
+    } catch (IllegalArgumentException e) {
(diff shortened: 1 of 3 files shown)


integration-tests/src/test/java/com/example/IntegrationTest.java
```diff
@@ -107,6 +107,20 @@ public void bindElementsIntoSetWrongReturn() {
...
+  @IgnoreCodegen
+  public void bindElementsIntoSetGenericWrongReturn() {
+    try {
+      backend.create(BindsElementsIntoSetGenericWrongReturn.class);
+      fail();
+    } catch (IllegalArgumentException e) {
+      assertThat(e)
+          .hasMessageThat()
+          .isEqualTo(
+              "@Binds methods must return a primitive, an array, a type variable, or a declared type. Found java.util.Set<? extends java.lang.String>.");
```
JakeWharton: Can you break this string at or around 100 chars?

runningcode: Formatted!

PR #161, opened 2019-08-06
Extract reflective parsing to Scope callers

JakeWharton:
Scope should not know about modules or dependency objects.


JakeWharton:
This is the first step in extracting it into a separate module which will help self-document boundaries (since it won't depend on Dagger). In that move types like `Key`, `Scope`, and `Binding` will become public and need to gain significant documentation.

PR #129, opened 2019-07-01
Allow nested generic types

runningcode:
Dont allow type variables in keys.

PR #167, opened 2019-08-07
Fix looking up generics when injecting fields in generic subclasses.

amatkovsky:
This pr adds a test to cover the case when there are @Inject annotated fields with type parameters. Real world example: a class `BaseFragment<P extends Presenter>` with a field `P presenter`.
The fix included is just an ad-hoc solution to highlight the issue and help come up with proper fix.


runningcode:
Is this a duplicate of this? https://github.com/JakeWharton/dagger-reflect/pull/129

amatkovsky:
> Is this a duplicate of this? #129

I am not sure. It looks like your pr deals with generics when resolving jit bindings while this pr is about resolving bindings for the fields.
But I think your solution may be expanded to cover this case too.

PR #169, opened 2019-08-09
Allow multiple interfaces to request same dependency.

Laimiux:
The added test displays the missing behavior.

### DIFF
--- a/reflect/src/main/java/dagger/reflect/ReflectiveDependencyParser.java
+++ b/reflect/src/main/java/dagger/reflect/ReflectiveDependencyParser.java
@@ -6,11 +6,14 @@ import dagger.reflect.Binding.LinkedBinding;
 import java.lang.annotation.Annotation;
 import java.lang.reflect.Method;
 import java.lang.reflect.Type;
+import java.util.LinkedHashSet;
+import java.util.Set;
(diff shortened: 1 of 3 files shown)


reflect/src/main/java/dagger/reflect/ReflectiveDependencyParser.java
```diff
@@ -21,9 +24,11 @@ static void parse(Class<?> cls, Object instance, Scope.Builder scopeBuilder) {
         Type type = method.getGenericReturnType();
         Key key = Key.of(qualifier, type);
 
-        Binding binding = new LinkedProvidesBinding<>(instance, method, NO_BINDINGS);
-
-        scopeBuilder.addBinding(key, binding);
+        if (!alreadySeen.contains(key)) {
```
JakeWharton: super nit / pet peeve:

you can do
```java
if (alreadySeen.add(key)) {
  // ...
}
```
to avoid having to do double hashing / lookup.

Laimiux: Done 👍

JakeWharton:
Thanks! Good find.

PR #183, opened 2019-12-09
Add generated annotation when possible

### LINKED ISSUE #172: Use @Generated annotation in the generated code
Creating this as a discussion topic - should the generated code have `@Generated` annotation or it doesn't bring any benefit.

JakeWharton:
Closes #172


reflect-compiler/src/main/java/dagger/reflect/compiler/DaggerReflectCompiler.java
```diff
@@ -48,12 +50,27 @@
...
+  private @Nullable AnnotationSpec generatedAnnotation;
 
   @Override
   public synchronized void init(ProcessingEnvironment processingEnv) {
     super.init(processingEnv);
     filer = processingEnv.getFiler();
     messager = processingEnv.getMessager();
+
+    String generatedAnnotationName =
+        processingEnv.getSourceVersion().compareTo(RELEASE_8) <= 0
```
runningcode: just curious, how come we check the java version number here in order to determine what annotation to use, but in the test we check if the class is available?

JakeWharton: Neither annotation exists in `android.jar`. So even if you are targeting <=8 or >8, the annotation might not actually be available which would lead the generated code to fail to compile.

runningcode: doesn't that mean that we should check its availability in both cases instead of the java version number?

JakeWharton: The version indicates which annotation will be correct for the generated code as determined by the user's build configuration. The availability of the annotation is a secondary, Android-only thing. It still might actually cause problems since you could be using Dagger in Java module where it seems like the annotation is available but when you consume it from an Android module the annotation will be missing and tools like R8/ProGuard will actually fail. In that case you have to pull in a dependency which contains the annotation.

PR #189, opened 2020-04-02
reflect-lint: Support Kotlin annotations with default retention

PhilGlass:
The [default retention for annotations written in Kotlin is RUNTIME](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.annotation/-retention/-init-.html), but `dagger-reflect-lint` currently flags any Kotlin annotation without an explicit `@Retention(AnnotationRetention.RUNTIME)` as an error.


TWiStErRob:
Based on the kdoc you linked to me it looks like you still must put `@Retention` so just `annotation class X` doesn't have RUNTIME retention. However your claim is correct in the end, but the relevant doc here is [AnnotationRetention.RUNTIME](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.annotation/-annotation-retention/-r-u-n-t-i-m-e.html) and the [annotation design doc](https://github.com/JetBrains/kotlin/blob/master/spec-docs/annotation-options.md#retention)

PhilGlass:
@TWiStErRob You're right, that's the wrong documentation link. It was late!

I found another issue with `AnnotationRetention.BINARY` support after running this check over our codebase - I've pushed another commit that fixes it.

PR #198, opened 2020-08-12
Do not eagerly link providers in a map binding

JakeWharton:
Previously we linked them when the map was linked. But we only need the map to be linked to capture the scope. The map's providers of values should still be late-linked upon first access.

Closes #175. cc @alexjlockwood

Issue #175, opened 2019-10-07
dagger-reflect does not resolve Provider/Lazy indirection based dependency cycles


matejdro:
Running following code:

```kotlin
@Singleton
@Component(modules = [MainModule::class])
interface MainComponent {
    fun provideFactory(): TestFactory
}

@Module
abstract class MainModule {
    @Binds
    @IntoMap
    @ClassKey(TestClassA::class)
    abstract fun bindA(testClassA: TestClassA): Any
}

@Singleton
class TestFactory @Inject constructor(
    private val providers: Map<@JvmSuppressWildcards Class<*>, @JvmSuppressWildcards Provider<Any>>
) {
    fun createA(): TestClassA {
        @Suppress("UNCHECKED_CAST")
        return (providers[TestClassA::class.java] as Provider<TestClassA>).get()
    }
}

class TestClassA @Inject constructor(val factory: TestFactory)

fun main() {
    Dagger.create(MainComponent::class.java)
        .provideFactory()
}
```

With:

```
implementation 'com.google.dagger:dagger:2.24'

// Snapshot fetched on Mon Oct 7 2019 8:00:00 UTC>
implementation 'com.jakewharton.dagger:dagger-reflect:0.2.0-SNAPSHOT'
```

Will result in a following crash:

```
Exception in thread "main" java.lang.IllegalStateException: Dependency cycle for test.TestFactory
 * Requested: test.TestFactory
     from dagger.reflect.UnlinkedScopedBinding@66cd51c3
 * Requested: java.util.Map<java.lang.Class<?>, javax.inject.Provider<java.lang.Object>>
     from Map{class test.TestClassA=@Binds[test.MainModule.bindA(…)]}
 * Requested: test.TestClassA
     from @Inject[test.TestClassA.<init>(…)]
 * Requested: test.TestFactory
     which forms a cycle.
	at dagger.reflect.Linker.failure(Linker.java:45)
	at dagger.reflect.Linker.link(Linker.java:32)
	at dagger.reflect.Scope.link(Scope.java:145)
	at dagger.reflect.Scope.findExistingBinding(Scope.java:103)
	at dagger.reflect.Scope.findBinding(Scope.java:74)
	at dagger.reflect.Linker.find(Linker.java:27)
	at dagger.reflect.Linker.get(Linker.java:18)
	at dagger.reflect.UnlinkedJustInTimeBinding.link(UnlinkedJustInTimeBinding.java:38)
[...]

Issue #4, opened 2019-01-31
Support for DaggerComponent.builder()


TKolbusz:
Hi,

a lot of us are probably using Dagger modules with constructor parameters. It would be great if you could use such a construction in dagger-reflect.

```
DaggerXComponent.builder()
                .xModule(new XModule(this))
                .build();
```

Thanks.

JakeWharton:
This is already tracked with a failing integration test. PR welcome, but in general you should refactor to `@BindsInstance` and stateless modules.

ColtonIdle:
@JakeWharton can you clarify why we should be using stateless modules?

JakeWharton:
They're faster and optimize better. If you're tempted to make stateful modules just put the state into the graph directly and use stateless modules instead.

TWiStErRob:
@JakeWharton So does this mean there's no plan to support `.builder()` unless someone creates a PR for it? In that case #191 is a dupe.

JakeWharton:
It would be great not to support it, yes. I don't want to duplicate the logic of Dagger for stateful module awareness and the associated builder logic.

PR #200, opened 2020-08-20
Creates cycle indirection scenarios for tests

gfreivasc:
Creates `IndirectionCycle` with regular provided dependencies,`BindsIndirectionCycle` for bound dependencies indirection, and `MultibindProviderMapIndirectionCycle` for the scenario pinpointed in issue #175 and resolved by #198.


gfreivasc:
I think we ran into a conflict here, but I managed to clean up branch history.

TWiStErRob:
There's a cycle test in upstream as well:
https://github.com/JakeWharton/dagger-reflect/blob/dec8f8f75256fe70d2fbd86d671b6723de562337/integration-tests/upstream/build.gradle#L21-L23
But I guess this will never execute because https://github.com/JakeWharton/dagger-reflect/issues/4#issuecomment-646408154

TWiStErRob:
What do you think about testing cycles and checking that the instances actually end up the same when scoped?

gfreivasc:
Quickly looking at it, it seems that perhaps some of these tests are okay to run. I'll take a look, since upstream currently covers two of the three tests.

gfreivasc:
I think the test scenario and tests suffice, right?

PR #203, opened 2020-09-11
Fix injection of MembersInjector

quiro91:


JakeWharton:
Thanks for sending a PR!

Your changes are functional, but I've decided to implement this a bit differently. I want to cache the `MembersInjector` that is created in the graph itself so it's only created once rather than creating one each time it's requested.

That implementation is in #205.

quiro91:
No problem, makes sense

PR #196, opened 2020-06-26
Fix unlinked Android injectors factories

crgarridos:
This commit fixes #174

Issue #174, opened 2019-10-02
ContributesAndroidInjector doesn't work for Fragment inside an Activity


kokeroulis:
If you have a nested `ContributesAndroidInjector` then it doesn't work properly. Dagger complains about a missing injector factory. Dagger with code generation works fine, only dagger reflect is crashing.

```kotlin
@Module
public interface ExampleActivityModule {

    @ContributesAndroidInjector
    ExampleFragment fragment();
}
```

Full Example [here](https://github.com/kokeroulis/dagger-reflect/tree/atsiap/module_issue_include) https://github.com/kokeroulis/dagger-reflect/commit/adbd4ab65ffd52847c68c5967d82f6454ba25090

**stacktrace**

```
E  java.lang.RuntimeException: Unable to start activity ComponentInfo{com.example/com.example.ExampleActivity}: java.lang.IllegalArgumentException: No injector factory bound for Class<com.example.ExampleFragment>
E      at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:2913)
E      at android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:3048)
E      at android.app.servertransaction.LaunchActivityItem.execute(LaunchActivityItem.java:78)
E      at android.app.servertransaction.TransactionExecutor.executeCallbacks(TransactionExecutor.java:108)
E      at android.app.servertransaction.TransactionExecutor.execute(TransactionExecutor.java:68)
E      at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1808)
E      at android.os.Handler.dispatchMessage(Handler.java:106)
E      at android.os.Looper.loop(Looper.java:193)
E      at android.app.ActivityThread.main(ActivityThread.java:6669)
E      at java.lang.reflect.Method.git invoke(Native Method)
E      at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:493)
E      at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:858)
E  Caused by: java.lang.IllegalArgumentException: No injector factory bound for Class<com.example.ExampleFragment>
E      at dagger.android.DispatchingAndroidInjector.inject(DispatchingAndroidInjector.java:136)
E      at dagger.android.AndroidInjection.inject(AndroidInjection.java:181)
E      at dagger.android.AndroidInjection.inject(AndroidInjection.java:91)
E      at com.example.ExampleFragment.onAttach(ExampleFragment.java:11)
[...]

PR #211, opened 2020-11-13
[Updated] Fix unlinked android injector factories

nodejsjs:
Updated PR for https://github.com/JakeWharton/dagger-reflect/pull/196 https://github.com/JakeWharton/dagger-reflect/issues/174
Credits to @crgarridos and @kokeroulis

* Added tests
* Formatted code using the googleJavaFormat task

### DIFF
--- a/integration-tests/android-has-interface/src/test/java/com/example/ExampleFragmentInjectionActivityTest.java
+++ b/integration-tests/android-has-interface/src/test/java/com/example/ExampleFragmentInjectionActivityTest.java
@@ -1,0 +1,27 @@
+package com.example;
+
+import androidx.test.core.app.ActivityScenario;
+
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.robolectric.RobolectricTestRunner;
+
+import static androidx.lifecycle.Lifecycle.State.CREATED;
+import static androidx.test.core.app.ActivityScenario.launch;
+import static com.google.common.truth.Truth.assertThat;
+
+@RunWith(RobolectricTestRunner.class) //
+public class ExampleFragmentInjectionActivityTest {
+    @Test
+    public void activityFragmentInjection() {
+        try (ActivityScenario<ExampleFragmentInjectionActivity> scenario = launch(ExampleFragmentInjectionActivity.class)) {
+            scenario.moveToState(CREATED);
+            scenario.onActivity(activity -> {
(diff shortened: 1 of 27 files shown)


integration-tests/android-has-interface/src/test/java/com/example/ExampleFragmentInjectionActivityTest.java
```diff
@@ -0,0 +1,27 @@
...
+            scenario.moveToState(CREATED);
+            scenario.onActivity(activity -> {
+                ExampleFragment fragment = (ExampleFragment) activity.getFragmentManager().findFragmentByTag("ExampleFragment");
+                assertThat(fragment.string).isEqualTo("Hello!");
+                assertThat(fragment.aLong).isEqualTo(10L);
+                assertThat(fragment.anInt).isEqualTo(20);
+            });
+        }
+    }
+}
```
nodejsjs: This test is failing when the task `:integration-tests:android-has-interface:testReflectDebugUnitTest` is run. But as a single test run from the IDE, it passes just fine. Any hints please on what might be going on here? @JakeWharton

TWiStErRob: It's probably because of moveToState. `launch` may have an optional parameter for initial state like [Fragment](https://github.com/TWiStErRob/repros/blob/master/espresso/fragment-lifecycle/src/androidTest/java/com/example/FragmentLifecycleTest.kt#L30),  even if it doesn't, the default initial state may be RESUMED. I think you are going to DESTROYED with this code. _Just an idea, not fully sure._

Also you may need to idle the looper, since Robolectric 4.4.

nodejsjs: Thanks for the pointers! I'm gonna investigate the Fragment/ActivityScenario apis more. I should add that it fails with the following error:

```
No injector factory bound for Class<com.example.ExampleFragment>
java.lang.IllegalArgumentException: No injector factory bound for Class<com.example.ExampleFragment>
	at dagger.android.DispatchingAndroidInjector.inject(DispatchingAndroidInjector.java:136)
	at dagger.android.support.AndroidSupportInjection.inject(AndroidSupportInjection.java:75)
	at dagger.android.support.AndroidSupportInjection.inject(AndroidSupportInjection.java:67)
	at com.example.ExampleFragment.onAttach(ExampleFragment.java:34)
```

Issue #213, opened 2021-03-06
Support for new Dagger AssistedInject


Dkhusainov:
Currently `dagger-reflect` doesn't work with the new AssistedInject feature.

JakeWharton:
It will be a while until I have time to work on this. If anyone wants to do any work in a PR it would be welcomed.

PR #214, opened 2021-03-09
AssistedInject implementation

Dkhusainov:
#213 

@JakeWharton There is not tag for latest dagger version in their repository... Otherwise `./gradlew build` works


JakeWharton:
Can you file a bug on Dagger for adding a tag as well? (If you haven't already)

integration-tests/upstream/test-source-fixed.gradle
```diff
@@ -0,0 +1,71 @@
...
+    }
+  }
+
+  String fixSource(String source) {
+    return source.readLines()
+      .collect { line -> fixSourceLine(line) }
+      .join("\r\n")
+  }
+
+  abstract String fixSourceLine(String sourceLine)
```
TWiStErRob: just curious: why split the source code rather than simply doing a replace on the whole file at once?

integration-tests/upstream/test-source-fixed.gradle
```diff
@@ -0,0 +1,71 @@
+//Tasks for "fixing" small things in upstream test sources and running them
+
+def output = file("$buildDir/test-source-fixed")
+
+sourceSets.test.java {
+  srcDir(output)
+}
+
+def fix = { String testClassName,
+            Class<TestSourceFix> taskClass ->
```
TWiStErRob, line 10: Strange to define a variable of a closure when it's just a function, how about
```suggestion
void fix(String testClassName, Class<TestSourceFix> taskClass) {
```
or better yet:
```suggestion
static void fix(Project project, String testClassName, Class<TestSourceFix> taskClass) {
```
just to need to declare which variable is captured from `project`.

TWiStErRob:
There's a tag for [2.33](https://github.com/google/dagger/releases/tag/dagger-2.33)

integration-tests/upstream/test-source-fixed.gradle
Dkhusainov: Can't think of a reason. Maybe it seemed more extendable

integration-tests/upstream/test-source-fixed.gradle
Dkhusainov: You can't capture local script variables in functions, so you'd have to pass each one(just `output` in this case) of them to the function in addition to passing `project`.

Dkhusainov:
> There's a tag for [2.33](https://github.com/google/dagger/releases/tag/dagger-2.33)

I see. Had to recheckout the whole thing for them to appear in my repo.

integration-tests/upstream/test-source-fixed.gradle
TWiStErRob: ... I'll never understand Groovy. 😅

TWiStErRob:
Submodules are fun... I think you would have needed to do a `git fetch` inside the submodule folder, not in this repo.

Dkhusainov:
Not all of them are simple visibility modifier changes. If we're fixing them, why bother going to dagger when we can just fix visibility too?

reflect/src/main/java/dagger/reflect/ReflectiveAssistedFactoryLookupFactory.java
```diff
@@ -0,0 +1,263 @@
...
+      return null; // AssistedFactory type cannot be qualified
+    }
+
+    Type type = key.type();
+
+    try {
+      return getJustInTimeAssistedFactoryLookup(type);
+    } catch (IllegalStateException t) {
+      throw new IllegalStateException(
+          String.format("Failed to create @AssistedFactory %s: \n%s", type, t.getMessage()), t);
```
abidaJustice, line 36: I detect that this code is problematic. According to the [Bad practice (BAD_PRACTICE)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#bad-practice-bad-practice), [FS: Format string should use %n rather than n (VA_FORMAT_STRING_USES_NEWLINE)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#fs-format-string-should-use-n-rather-than-n-va-format-string-uses-newline).
This format string includes a newline character (\n). In format strings, it is generally  preferable to use %n, which will produce the platform-specific line separator.

Dkhusainov:
@JakeWharton is this project dead?

JakeWharton:
No, but it's not a high priority. I'm not paid to work on it. Feel free to have your company sponsor me on GitHub if it's important to you.

Issue #202, opened 2020-09-08
dagger-android crash in DaggerApplication


realdadfish:
* Have your main application class derive from `dagger.android.DaggerApplication`
* Implement `applicationInjector()` to return the `AndroidInjector`
* Partial reflection setup
* Dagger: 2.28.3, Delect 0.3.0 (Dagger-Reflect-latest)

```
java.lang.RuntimeException: Unable to create application my.application.MyApplication: java.lang.IllegalStateException: The AndroidInjector returned from applicationInjector() did not inject the DaggerApplication
      at android.app.ActivityThread.handleBindApplication(ActivityThread.java:6715)
      at android.app.ActivityThread.access$1300(ActivityThread.java:226)
      at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1898)
      at android.os.Handler.dispatchMessage(Handler.java:107)
      at android.os.Looper.loop(Looper.java:214)
      at android.app.ActivityThread.main(ActivityThread.java:7615)
      at java.lang.reflect.Method.invoke(Native Method)
      at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:492)
      at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:964)
Caused by: java.lang.IllegalStateException: The AndroidInjector returned from applicationInjector() did not inject the DaggerApplication
      at dagger.android.DaggerApplication.injectIfNecessary(DaggerApplication.java:65)
      at dagger.android.DaggerApplication.onCreate(DaggerApplication.java:38)
      at my.application.MyApplication.onCreate(MyApplication.kt:46)
      at android.app.Instrumentation.callApplicationOnCreate(Instrumentation.java:1190)
      at android.app.ActivityThread.handleBindApplication(ActivityThread.java:6710)
      ... 8 more
```

JakeWharton:
Presumably because we don't support injecting a `MembersInjector`.

realdadfish:
Issue is that `AndroidInjector` has a `void inject(T instance);` and therefor does not enforce a specific `void inject(MyApplication instance)` method in the application component (which derives from said interface). `dagger-reflect` now reads this method's argument as `Object` and can't find the factory that provides injectees for this type.

I refactored the code to not use `DaggerApplication` and replaced the usage of `AndroidInjector` by a specific `inject(...)` method and that made this issue go away.

JakeWharton:
[...]

PR #216, opened 2021-03-26
Fix injection when using generics

AlexDochioiu:
Currently, dagger-reflect doesn't work with injectors interfaces relying on generics, such as:

```kotlin
interface Injector<T> {
    fun inject(instance: T)
}

@Component
interface MainActivityComponent : Injector<MainActivity>
```

This PR would fix this issue.


AlexDochioiu:
Haven't tried it but this might also fix #202

